PR #106, opened 2013-12-23
Zypper support

keremc:
This adds support for zypper, the package manager of openSUSE.

I have tested it on openSUSE 13.1.
All tests pass. (But first you'll need to install GNU patch and shunit2 and then set `SHUNIT2`)
(`sudo zypper in patch shunit2`
`export SHUNIT2=/usr/share/shunit2/src/shunit2`)

Rubinius fails to build. (at least on my PC, can someone check if it works for them?)
On my PC, Rubinius' build script complained about bundle being missing. After discovering that `install_optional_deps()` was for some reason failing to install bundle, I installed it manually with zypper.
(`sudo zypper in rubygem-bundler`)
However even after that Rubinius still refused to build, giving me this error:
`g++: error: unrecognized command line option ‘-Wcovered-switch-default’`

Ruby (tested 2.0.0-p353) and jRuby (tested 1.7.9), on the other hand, build successfully.


havenwood:
@keremc Thanks!

PR #131, opened 2014-03-07
update package name for apt jre

cpuguy83:
openjdk-7-jre installs all kinds of GUI stuff.
This doesn't work in Docker, use the headless package instead.

### DIFF
--- a/share/ruby-install/jruby/dependencies.txt
+++ b/share/ruby-install/jruby/dependencies.txt
@@ -1,3 +1,3 @@
-apt: openjdk-7-jdk
+apt: openjdk-7-jre-headless
 yum: java-1.7.0-openjdk
 pacman: jre7-openjdk


stefanhorning:
Would also like to see this merged. Why install a load of garbage first and then leave it to the user to clean up?
In my opinion this should install dependencies (as the name suggests) to run jruby not anything else that's somehow related to java (which I guess it not the task of ruby-install)...
BTW this is the list of packages that get installed on a Ubuntu server as a result of this https://gist.github.com/stefanhorning/49b5a75fea98ce802632

Thanks for your consideration.

postmodern:
@stefanhorning that's a convincing argument. However, I worry it would unfairly punish desktop jruby users. Perhaps we need to add a configuration option to switch between GUI and headless JREs?

cpuguy83:
Closing as this isn't really needed these days.

PR #138, opened 2014-04-02
Allow installing MRI from a binary mirror

sarahhodne:
This downloads and installs a precompiled version of MRI from an RVM-compatible mirror if `--binary` is passed

The system auto-detection is a huge mess, and probably violating the style guide (several lines longer than 80 characters). It's mostly lifted from RVM's own system detection. I don't know if you want this somewhere else (if so, where?) or if you want me to shorten the number of supported systems, just let me know and I can modify.

As for the long lines, should I just "wrap" them?

This currently just fails if `--binary` is passed and no binary build was found, since that was the easiest to implement.

### DIFF
--- a/bin/ruby-install
+++ b/bin/ruby-install
@@ -35,9 +35,13 @@ extract_ruby   || fail "Unpacking of $ruby_archive failed!"
 download_patches || fail "Fetching patches $patches failed!"
 apply_patches  || fail "Patching $ruby $ruby_version failed!"
 cd "$src_dir/$ruby_src_dir"
-configure_ruby || fail "Configuration of $ruby $ruby_version failed!"
-clean_ruby     || fail "Cleaning $ruby $ruby_version failed!"
-compile_ruby   || fail "Compiling $ruby $ruby_version failed!"
+
+if [[ ! $binary_install -eq 1 ]]; then
+	configure_ruby || fail "Configuration of $ruby $ruby_version failed!"
+	clean_ruby     || fail "Cleaning $ruby $ruby_version failed!"
+	compile_ruby   || fail "Compiling $ruby $ruby_version failed!"
+fi
+
 install_ruby   || fail "Installation of $ruby $ruby_version failed!"
 post_install   || fail "Post-install tasks failed!"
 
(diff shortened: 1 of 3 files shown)


dukz:
Hello, any chance that this will still be merged?

trammel:
Has there been any further thoughts on this? While the package manager solution might be preferable, the rvm provided binaries exist now and are usable.

airhorns:
Anything we can do to push this forward? I agree that package managers are always going to be better at this, however, they won't know about or play nice with `ruby-install`'s existing rubies. If you want to use the same system to manage rubies from source and rubies from binary in the same folder hierarchy (like as is required for `rbenv` or `chruby`), then you're a bit sunk. Can we add support for installing from binary with a `--i-know-what-im-doing-this-is-the-right-arch` flag or what have you to at least let users do it manually?

airhorns:
@postmodern sorry to bother on this again, would you mind at least sharing if you've committed to the idea of package managers installing binaries? If not, can I do anything to help push this PR through? Thanks for any pointers you can give!

postmodern:
@airhorns yes, I'm leaning more towards allowing users to generate their own packages. It would be possible to generate the necessary rpm/deb spec files using the information already available in ruby-install (dependencies and versions/checksums). Then we could generate ruby packages that install into `/opt/rubies/`.

tisba:
tiny attempted to revive this topic…

ruby-install is very nice in terms of flexibility, but being able to have pre-build binaries would make things soooo much faster and nicer.

Is there anything one can do? Like testing?

k3rni:
Bumping this again - recently I had to install ruby 2.3.8 which proved very hard due to openssl 1.1, and no amount of `--with-openssl-dir` could change that. I turned to [travis-built rubies](http://rubies.travis-ci.org/), and the nice thing was that it sufficed to unpack the tarball into `.rubies` and chruby was instantly happy. So I checked whether this was a supported thing, and turns out it's very much stalled.

sarahhodne:
(Closing this, since I don't really use Ruby much anymore, and don't think I have time to finish this PR)

PR #183, opened 2014-10-31
Install libedit-dev via apt

jc00ke:
LLVM 3.5 will add `-ledit` to the `config[:llvm_ldflags]` option.


postmodern:
libedit-dev(el) is now installed for apt, dnf, yum, and pacman.

PR #202, opened 2015-02-09
add rbx-2.5.{0,1,2}

Mic92:


havenwood:
Thank you for your pull request! Updates for versions and checksums are actually now commited to the [ruby-versions](https://github.com/postmodern/ruby-versions) repository and pulled from there. These rbx versions have already been added to [ruby-versions] but pull requests there are very much welcome.

PR #208, opened 2015-03-08
Update README installation instructions to download latest

CADBOT:
After following the old installation scripts, I was confused as to why I couldn't install ruby 2.2.0, and I imagine others have had this confusion. Updating the readme to point to the newest version would be helpful.


postmodern:
Ended up skipping 0.5.1 and going directly to 0.6.0.

PR #211, opened 2015-04-07
detect correct package manager on archlinux based distros

Azulinho:
on archlinux based distros, one can have multiple package managers
installed. There are AUR pkgs available for homebrew, yum, apt and
so on.
This allows for someone in a archlinux box to build packages for others
distributions, or to simply install them when they don't conflict
with OS files, examples are packages that install into /opt/<pkg>

The previous detection method would look for apt, yum and would set
the packager command to match whatever it found first.
My archlinux box has yum, brew, however the correct package manager
is pacman. A quickfix is to search for pacman first and then the
others.
Usually EL or Debian based boxes do not have 'brew', 'pacman' or
other package managers installed or available as a package.

*Works on my machine*

### DIFF
--- a/share/ruby-install/ruby-install.sh
+++ b/share/ruby-install/ruby-install.sh
@@ -15,13 +15,12 @@ make_opts=()
 #
 # Auto-detect the package manager.
 #
-if   command -v pacman	>/dev/null; then package_manager="pacman"
-elif command -v yum	>/dev/null; then package_manager="yum"
+if   command -v pacman  >/dev/null; then package_manager="pacman"
+elif command -v yum     >/dev/null; then package_manager="yum"
 elif command -v apt-get >/dev/null; then package_manager="apt"
-elif command -v port	>/dev/null; then package_manager="port"
-elif command -v brew	>/dev/null; then package_manager="brew"
+elif command -v port    >/dev/null; then package_manager="port"
+elif command -v brew    >/dev/null; then package_manager="brew"
 fi
-
 #
 # Auto-detect the downloader.
 #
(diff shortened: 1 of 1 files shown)


postmodern:
You might want to redo this commit. You're text editor replaced the hard-tabs with spaces. Bash, C, C++, Java should be hard-tabbed.

postmodern:
Perhaps we should use an alternate way of determine what type of system we are on, and then select the primary package manager?

Azulinho:
that's me and vim :retab which I happen to do without even thinking about it anymore.
Happy to re-introduce the tabs if you want me to. However I disagree that bash should be hard-tabbed, there isn't a proper official bash style guide but a few out there, possibly the most relevant is the one used by google which actually states always use 'spaces'.

https://google-styleguide.googlecode.com/svn/trunk/shell.xml#Indentation

About finding out the OS you could use a mix of $OSTYPE, and lsb_release that will give you if its a linux/osx box and which distro it is.

\>echo $OSTYPE
linux-gnu

\>lsb_release -i
Distributor ID:	ManjaroLinux

and this:
http://unix.stackexchange.com/questions/6345/how-can-i-get-distribution-name-and-version-number-in-a-simple-shell-script

postmodern:
@Azulinho yes please use tabs. Tabs seem to be the defacto standard. It's more of a "when in rome" kind of thing.

I will accept the immediate fix of checking for `pacman` first. Going forward, we could use this [beefy system detection code](https://github.com/postmodern/chruby/blob/c1d9daec77f9def1ad51ddad3ddcdfb2ac12a489/test/setup#L39-L150) we extracted from RVM.

Azulinho:
@postmodern just rebased your latest changes into my branch and pushed the fix for archlinux

postmodern:
Manually cherry-picked it in. Thanks! Also good catch on moving `yum` above `apt-get`, apparently you can `yum install apt` on Fedora.

Issue #215, opened 2015-04-20
Ruby 2.2.2 support any time soon?


dakull:

aprescott:
Might want to check out #214, #212 and the [ruby-versions](https://github.com/postmodern/ruby-versions).

dakull:
@aprescott thanks

JuanitoFatas:
@dakull Did you accidentally reopened this issue?

dakull:
seems so

PR #212, opened 2015-04-13
Add Ruby 2.2.2

nlindley:

PR #216, opened 2015-04-22
added support for Ruby 2.2.2 - #215

dakull:


aprescott:
Please read https://github.com/postmodern/ruby-install/pull/212#issuecomment-93062156 — this PR is in the wrong place and it looks like the ruby-versions repo was already updated.

dakull:
@aprescott  ok / thanks - sorry for the confusion

aprescott:
No worries, it's confusing. :)

postmodern:
How can we make ruby-versions more visible? I mentioned ruby-versions in the README and CONTRIBUTING.md.

aprescott:
@postmodern I think if it's in the README, it needs to be more prominent, as a notice to people who are about to try and update the versions, with an explicit call out for how to do that.

But ultimately, people may still want to go into `share/` (for checksums, versions, ...) and update files that should come from ruby-versions. If those files _must_ stick around in ruby-install, then it seems like it's wise to mention, somewhere, that they are not to be edited in ruby-install.

Since adding the idea of comments to the text files is probably out of the question (horrible backwards-compatibility issues), maybe just add a `_README` file at the top of `share/ruby-install/ruby/`, `share/ruby-install/ruby/versions.txt`, etc., with a big notice about updates to those files belong in ruby-versions?

dakull:
@postmodern @aprescott I agree

>Contains a list of known versions and checksums from ruby-versions.

vs.

>Does not require updating every time a new Ruby version comes out.

Maybe somthing using git submodules would work? (I do realize this creates a hard dependency on git)

aprescott:
Why keep the files in version control at all? Why not just `.gitignore` them and have the files generated at build time?

postmodern:
@aprescott that's a good idea and we already have the make task to create those files. Unfortunately, we're stuck using github's release system which creates the release archives from the git tags. Once I implement the caching/update system, I will remove the files entirely, and allow `ruby-install` to auto-download them.

PR #217, opened 2015-04-30
pacman: use llvm35 dependency for rbx install

ianks:
Rbx needs llvm 3.0 - 3.5, so we lock in the pacman dependemcy on the llvm35 package.


postmodern:
Sounds good. Wondering if we should do the same for apt-get? Debian/Ubuntu have llvm-3.1 and 3.5 packages.

PR #221, opened 2015-05-09
Explicitly install llvm-3.5 via apt

jc00ke:
3.6 doesn't work at the moment, which is the latest in Ubuntu 15.04. Need to lock it to 3.5 for now.


postmodern:
According to [packages.ubuntu.com](http://packages.ubuntu.com/vivid/llvm-dev) llvm-dev only maps to 3.6 on vivid, and llvm-3.5 is only available on trusty and higher. What is the oldest version of Ubuntu Rubinius supports?

jc00ke:
Probably 12.04

postmodern:
Manually merged the patch. :+1: May also want to update your fork.

PR #240, opened 2015-10-27
Add support for Fink on Mac OS X

fingolfin:
This PR adds support for the Fink package manager on Mac OS X.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Installs [Ruby], [JRuby], [Rubinius], [MagLev] or [mruby].
   * [pacman]
   * [macports]
   * [brew]
+  * [fink]
 * Has tests.
 
 ## Anti-Features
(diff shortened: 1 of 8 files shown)


fingolfin:
So I rebased this on latest master. Is there anything I can do to get this merged? No feedback at all for 8 months is a bit discouraging :-(.

For reference, Fink was one of the first (if not the first?) package managers for Mac OS X. It predates MacPorts. And while it is by far not as popular as Homebrew these days, there are still quite some people (esp. in academia) using it.

havenwood:
Thanks for the patch!

When we discussed OS X package managers a few years ago the thinking was to draw the line at Fink. If there are lots of folk using it then maybe that can be revisited.

fingolfin:
First of, thanks for the reply!

Well, how exactly do you quantify "lots of folks"? And what would be an acceptable "proof" that should a number of users is reached?

Also,, given how trivial this patch is, I don't quite understand the rationale for not including it based on user numbers -- alas, that's just me, and you most certainly don't need to justify your position to me -- I am simply wondering :-)

fingolfin:
Perhaps one can read that discussion somewhere (that might explain the reasons)? Or was it a private discussion?

havenwood:
A few years ago there was a public discussion about Fink in the #chruby channel on Freenode.net as well as a few follow-up tweets. I don't have the logs handy but please stop by the #chruby IRC channel if you'd like to discuss.

Issue #227, opened 2015-07-12
Support dnf package manager


anonova:
Fedora 22 [replaced yum with dnf](http://dnf.baseurl.org/2015/05/11/yum-is-dead-long-live-dnf/), which is supposed to be a drop-in replacement for yum. This means the package names currently described for yum are the same. The interface is mostly the same as well.

Note that this can be a low priority issue since Fedora currently has a shim for yum to redirect to dnf. Even though it will print a deprecation warning, ruby-install still works as expected on F22. It would just be nice not to see such warnings.

```
Yum command has been deprecated, redirecting to '/usr/bin/dnf install -y gcc automake zlib-devel libyaml-devel openssl-devel gdbm-devel readline-devel ncurses-devel libffi-devel'.
See 'man dnf' and 'man yum2dnf' for more information.
To transfer transaction metadata from yum to DNF, run:
'dnf install python-dnf-plugins-extras-migrate && dnf-2 migrate'
```

postmodern:
Implemented in 4560b6dbe39cc1c2ff23bc04cbfa380cd81cdab1.

PR #242, opened 2015-11-07
add support for the dnf package manager.

sockmonk:
dnf is supposed to eventually replace yum, but can be used
as a drop-in replacement.


sockmonk:
This is intended to address https://github.com/postmodern/ruby-install/issues/227. 
I forgot to look for existing pull requests; I think I wound up duplicating @FreedomBen's pull request.

postmodern:
Already merged, thanks!

PR #244, opened 2015-11-09
sudo optionally reads password from standard input

craigfurman:
The `-S` flag allows `sudo` to read the password from standard input if there is any available. If there are nothing to read from stdin, ruby-install will not behave any differently to how it does today.

This is useful for running ruby-install as non-root users in a non-interactive way, for example in [my workstation ansible playbook](https://github.com/craigfurman/ansible-home/blob/master/roles/common/tasks/main.yml#L141).

Cheers!


postmodern:
Hmm, perhaps we should check if we're running in a real terminal, and fallback to `-S`?

craigfurman:
Do you mean check if `ruby-install` is running on the right of a pipe, and use `sudo -S` if we are? The `-S` flag doesn't *require* stdin to have content ready, if it is empty is prompts for password as normal. Is there another reason why we would want to do this sort of check that I've missed?

postmodern:
I was thinking of `[[ -t $0 ]]` or `[[ "$-" == *i* ]]` to test if we actually have an interactive terminal vs. a non-interactive ssh session.

craigfurman:
I've implemented something to that effect, but since ruby-install calls some scripts `$0` is never a terminal, so $sudo is set to "sudo -S" every time. I'll look into detecting whether a session is interactive or not later.

craigfurman:
Sorry for leaving this stale for so long. I no longer use this project, so I won't be picking this back up 😅

PR #251, opened 2015-12-29
functions.sh: cleanup_source: remove src_dir if empty

### LINKED ISSUE #239: `--cleanup` should remove `src` directory
From https://github.com/postmodern/ruby-install/issues/236#issuecomment-146638902, regarding `cleanup`:

I was under the impression the `src` dir would be removed, but that appears to not be the case. Why not, though? I understand you might not want to blindly delete the directory in case it already existed, but it seems to me a `cleanup` option should indeed get rid of it if it wasn’t there before. I can see a few options:
- `rmdir` it. This way, the directory will only be gone if it is empty.
- When first creating the directory, check if it already existed. If it didn’t, make a note of it and delete it after everything is done (only if `--cleanup` was given, naturally).

If any of those sounds good to you, I’d be glad to look through the code and submit a PR. The first solution is easy enough and I already found where to do it, whereas for the second I’ll have to read the code a bit more.

vitorgalvao:
Closes #239.

### DIFF
--- a/share/ruby-install/functions.sh
+++ b/share/ruby-install/functions.sh
@@ -133,4 +133,7 @@ function cleanup_source() {
 
 	log "Removing $src_dir/$ruby_dir_name ..."
 	rm -rf "$src_dir/$ruby_dir_name" || return $?
+
+	log "Removing $src_dir ..."
+	rmdir "$src_dir" || return $?
 }


vitorgalvao:
Anything different you’d like to see here, @postmodern?

postmodern:
I worry whether some users might alias `rmdir` to something else... Also, I'm planning on downloading the files to `~/.cache/` instead of `~/src/`.

vitorgalvao:
> I worry whether some users might alias `rmdir` to something else...

What do you mean? [`rmdir` is a standard unix command](https://en.wikipedia.org/wiki/Rmdir). Worrying users alias it to something else is the same as worrying they alias `rm` or `cp` or `mv` to something else. It makes no sense. If you alias a standard command to something non-standard, it’s your fault and there’s nothing you as a developer can do about that.

> Also, I'm planning on downloading the files to `~/.cache/` instead of `~/src/`.

So just change it when/if you do it. I don’t see how that prevents this PR.

postmodern:
Many standard unix commands are routinely re-aliased by distros, 3rd-party shell configs, and users. Accidentally nuking someone's `~/src/` is not ideal. Once ruby-install fully switches over to `~/.cache/ruby-install/`, we can just leave the directory there, and only worry about deleting downloaded files.

vitorgalvao:
> Many standard unix commands are routinely re-aliased by distros, 3rd-party shell configs, and users. 

Even if that is the case, there’s also the fact that shell scripts don’t use your aliases (unless you specifically tell them to). If they did, you’d never be able to run a shell script with peace of mind.

> Accidentally nuking someone's ~/src/ is not ideal.

So that will not happen.

> Once ruby-install fully switches over to `~/.cache/ruby-install/`, we can just leave the directory there, and only worry about deleting downloaded files.

So closing both this and #239 seems to be like the best course of action, no? Why keep it open?

vitorgalvao:
Closing as part of personal cleanup of old PRs and issues. I don’t think this will ever be merged and I have long stopped using the tool, so no point to inflating the number of PRs.

PR #260, opened 2016-02-03
Run brew via absolute path

ithinkihaveacat:
`sudo -u "$brew_owner" brew` fails if brew isn't in the system PATH. (At least on OS X, sudo seems to have special treatment of PATH; even `-E` doesn't preserve it.)

To work around this, figure out the full path to brew before the sudo call.

### DIFF
--- a/share/ruby-install/util.sh
+++ b/share/ruby-install/util.sh
@@ -95,8 +95,8 @@ function install_packages()
 		port)   $sudo port install "$@" || return $?       ;;
 		brew)
 			local brew_owner="$(/usr/bin/stat -f %Su "$(command -v brew)")"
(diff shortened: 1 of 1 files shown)


havenwood:
I'm not able to reproduce `sudo -u` dropping env.If you do drop env, `which brew` won't locate brew - since `brew` won't be in your `PATH`.

havenwood:
Can you still reproduce this, @ithinkihaveacat? I can't reproduce the problem and the solution doesn't seem to me like it should work.

ithinkihaveacat:
Thanks for looking into this. I can't reproduce this now, so either the behaviour changed, or I was mistaken. I'll close this.

(The `$(which brew)` worked because it's executed when assembling arguments, before the `sudo` happens.)

PR #262, opened 2016-03-06
Add example of installing a remote binary build

donv:
This also illustrates how to install snapshots as an alternative to installing trunk/HEAD.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,10 @@ Install a Ruby from a mirror:
 
     $ ruby-install -M http://www.mirrorservice.org/sites/ftp.ruby-lang.org/pub/ruby ruby 2.0.0-p645
 
+Install a Ruby from a remote tarball:
+
+    $ ruby-install -u http://ci.jruby.org/snapshots/master/jruby-bin-9.1.0.0-SNAPSHOT.tar.gz jruby 9.1.0.0-SNAPSHOT
+
 Install a Ruby with a specific patch:
 
     $ ruby-install -p https://raw.github.com/gist/4136373/falcon-gc.diff ruby 1.9.3-p551
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Install a Ruby from a mirror:
 
 Install a Ruby from a remote tarball:
 
-    $ ruby-install -u http://ci.jruby.org/snapshots/master/jruby-bin-9.1.0.0-SNAPSHOT.tar.gz jruby 9.1.0.0-SNAPSHOT
+    $ ruby-install -u https://s3.amazonaws.com/jruby.org/downloads/9.0.5.0/jruby-bin-9.0.5.0.tar.gz jruby 9.0.5.0
 
 Install a Ruby with a specific patch:
 


postmodern:
I prefer to discourage users from installing from HEAD/trunk.

donv:
Yes, this is not installing from HEAD/trunk.  It is installing from a URL to a binary tarball.  I had to research a bit to find how it should be done with ruby-install, so I thought I'd share it with others :smile:

postmodern:
The URL does contain `snapshots/master`. Are you sure that isn't a nightly snapshot URL?

donv:
Ah , yes.  I probably misunderstood your meaning earlier.  Yes, the link points to a nightly build, built from HEAD of the master branch.

I thought you meant "install from HEAD/trunk source".

The link itself can be useful, but the main point is to display the use of the `-u` option.

I can change the link to a stable release tarball like https://s3.amazonaws.com/jruby.org/downloads/9.0.5.0/jruby-bin-9.0.5.0.tar.gz if you think that is better.

postmodern:
Yeah, definitely change it to a stable URL. :)

donv:
@postmodern OK, changed the URL to a stable release.  I still think a link to a snapshot makes more sense, but your are the boss :)

postmodern:
@donv hmm but now that is the same URL as the default jruby download URL. Perhaps we should make a Wiki page documenting how to install snapshots? I don't want to encourage installing trunk/HEAD/edge/nightly, but am not opposed to documenting it for those who are interested.

donv:
@postmodern Sounds good.  Where should we add that WIKI page?  Maybe something like this:  

https://github.com/postmodern/ruby-install/wiki/Install-a-remote-snapshot-tarball ?

Linked from the main WIKI page:  https://github.com/postmodern/ruby-install/wiki

donv:
Personally I use the -c flag when I download snapshots, since I most likely want a fresh one next time I want to install.  Should I add that to the example?

donv:
As a side note:  I experience that when I install a new JRuby snapshot over another, the new one does not take effect.  The version (revision and date) is not changed.  I have to manually `rm -rf` the directory under `.rubies` before installation to allow the new version to take effect.

Is that expected?

postmodern:
@donv that sounds like a bug. Appears we only `cp -R` the jruby archive to the installation directory.

postmodern:
I updated the formatting on the Wiki page. Think we can close this issue now.

donv:
@postmodern Regarding the `cp -R` bug, should I do something, or are you on it?  :smile:

postmodern:
@donv I thought about `rm -rf`ing the previous install, but that would cause havoc if a user was installing into `/usr/local`. Can you isolate which files are preventing the new jruby from being loaded? Are you also using NailGun or one of those JVM pre-loaders?

donv:
Not sure I understand your first question.  I would expect all files to be overwritten by the new install, not some files in particular.

Isn't the problem that `cp -R` does not overwrite files?

`rm -rf` works for my case :smile: 

I am not using NailGun or other JVM/Ruby/Rails preloaders.

postmodern:
@donv `cp -R` should overwrite the files.

PR #268, opened 2016-04-10
Avoid copying into subdirectory in reinstall

donv:
Fix for reinstalling a JRuby tarball.  The extracted directory was copied into an earlier install if present.

Came about after discussion in https://github.com/postmodern/ruby-install/pull/262#issuecomment-203615464 .

I have tested installing with no `~/.rubies`, no previous install, and with a previous install of the same JRuby version.

### DIFF
--- a/share/ruby-install/jruby/functions.sh
+++ b/share/ruby-install/jruby/functions.sh
@@ -11,7 +11,7 @@ ruby_url="${ruby_url:-$ruby_mirror/$ruby_version/$ruby_archive}"
 function install_ruby()
 {
 	log "Installing jruby $ruby_version ..."
-	cp -R "$src_dir/$ruby_dir_name" "$install_dir" || return $?
+	cp -R "$src_dir/$ruby_dir_name" "$rubies_dir/" || return $?
 }
 
 #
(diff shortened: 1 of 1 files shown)


donv:
Hi @postmodern !

Does this look OK to merge?

share/ruby-install/jruby/functions.sh
```diff
@@ -11,7 +11,7 @@ ruby_url="${ruby_url:-$ruby_mirror/$ruby_version/$ruby_archive}"
 function install_ruby()
 {
 	log "Installing jruby $ruby_version ..."
-	cp -R "$src_dir/$ruby_dir_name" "$install_dir" || return $?
+	cp -R "$src_dir/$ruby_dir_name" "$rubies_dir/" || return $?
```
postmodern: One minor issue here. Users can override `install_dir` via `--install-dir`.

postmodern: How about `cp -R "$src_dir/$ruby_dir_name/"* "$install_dir/"`

havenwood: Copying the contents of `$ruby_dir_name` to `$install_dir` seems like a nice solution. @donv ^

donv: Seems good to me.

donv:
I changed to copy the content to `install_dir`.

postmodern:
This might be fixed in c339bf6ca481ec6fac7fd18ec537f46bed0d1ec5 and 765498527dac7104b7ef31cc1907151059f85389.

donv:
Yeah, probably.

PR #271, opened 2016-05-25
Update apt ruby-dev in rbx dependencies.

zacts:
This updates the apt section of the rbx dependencies.txt to use a more generic ruby-dev, to allow for non-broken dependencies on Debian 8.0 Jessie and Ubuntu 16.04 LTS. (I have only tested this on Debian 8.0 Jessie via apt-get install ruby-dev (which falls back to the default ruby2.1-dev), but if you want I can test it on Debian Jessie and Ubuntu 16.04 vagrant images).


postmodern:
Ah this is a good point. Rubinius does require [ruby 2.0.0+](http://rubinius.com/doc/en/getting-started/requirements/). Ubuntu Precise (12.04LTS) still has their [ruby package](http://packages.ubuntu.com/precise/ruby-dev) pointing to ruby1.8, but does provide a [ruby1.9.1 package](http://packages.ubuntu.com/precise/ruby1.9.1). I believe Rubinius can still build with Ruby 1.9.1, but only by coincidence. Ubuntu trusty (14.04LTS) is the oldest version that provides a [ruby2.0 package](http://packages.ubuntu.com/trusty/ruby2.0). While Debian stable doesn't appear to have a ruby2.0 package, it's [ruby-dev package](https://packages.debian.org/stable/ruby-dev) is already at 2.1.

postmodern:
Fixed in 7f3e59f220d95d5a4279a4bae3158993fd5b9d1c.

PR #272, opened 2016-05-25
Update apt llvm-dev in rbx dependencies.

zacts:
This updates the apt section of the rbx dependencies.txt to use a more generic llvm-dev, to allow for non-broken dependencies on Debian 8.0 Jessie and Ubuntu 16.04 LTS. (I have only tested this on Debian 8.0 Jessie via apt-get install llvm-dev (which falls back to the default llvm-3.5-dev), but if you want I can test it on Debian Jessie and Ubuntu 16.04 vagrant images).


postmodern:
A common problem with Rubinius, is they often do not support newer or older versions of LLVM. `llvm-dev` may vary depending on which distro version you're running.

postmodern:
Fixed in 7f3e59f220d95d5a4279a4bae3158993fd5b9d1c.

PR #276, opened 2016-07-01
Upgrade openjdk to openjdk-8-jdk

ngpestelos:


postmodern:
[openjdk-8-jdk](http://packages.ubuntu.com/search?keywords=openjdk-8-jdk) was added in Ubuntu willy, which means this wouldn't work on Ubuntu vivid, trusty or precise. Is there a virtual package we could use to get the latest openjdk?

ngpestelos:
There is a [PPA](https://launchpad.net/~openjdk-r/+archive/ubuntu/ppa) for openjdk-8-jdk, but I don't know where to put this repository in the code. Should this be part of the `fetch` function?

jodosha:
👍  On Ubuntu 16.10 it's impossible to install JRuby with `ruby-install`.

---

@postmodern 

> openjdk-8-jdk was added in Ubuntu willy, which means this wouldn't work on Ubuntu vivid, trusty or precise.

Can't vidid/trusty/precise's users to upgrade to the newest JDK? I mean, it's a matter of following latest Java/JRuby improvements. WDYT? :)

deivid-rodriguez:
> :+1: On Ubuntu 16.10 it's impossible to install JRuby with ruby-install.

Just run into this myself. One workaround is to install the dependencies manually and then run `ruby-install` with the `--no-install-deps` flag.

havenwood:
#311 suggests `default-jdk` for apt, but there isn't an equivalent for pacman.

havenwood:
Closes #283.

havenwood:
The JRuby folk agree that this is the best course of action. We should also bump zypper to 8.

havenwood:
@ngpestelos Would you mind rebasing, bumping zypper and PRing to 0.7.0? Thank you for the PR!

havenwood:
Trusty Tahr is still missing openjdk-8-jdk, fwiw. I think at this point, if there must be a choice, it's better to support Xenial Xerus and later and have Trusty and earlier use ruby-install 0.6.0.

It seems this will continue to be an intermittent recurring issue as JRuby depends on higher versions of the JDK.

havenwood:
Closing since we've bumped to 8 on the 0.7.0 branch, with Zypper as well.

Thank you for the PR!! ✨

Issue #63, opened 2013-08-13
BSD `stat` and GNU `stat` take different options


paul:
When installing any ruby, I get:

```
» ruby-install ruby 2.0
>>> Installing ruby 2.0.0-p247 into /Users/rando/.rubies/ruby-2.0.0-p247 ...
>>> Installing dependencies for ruby 2.0.0-p247 ...
stat: cannot read file system information for ‘%Su’: No such file or directory
sudo: unknown user:   File: "/usr/local/bin/brew"
    ID: 100000400000011 Namelen: *       Type: hfs
Block size: 4096       Fundamental block size: 4096
Blocks: Total: 60987521   Free: 18827172   Available: 18763172
Inodes: Total: 60987519   Free: 18763172
!!! Installing dependencies failed!
```

This is because BSD `stat` (default on Darwin) takes a `-f "%Su"` option, while GNU coreutils `stat` uses `-c "%u"`

```
~  » stat --version
stat (GNU coreutils) 8.21
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Michael Meskes.
~  » stat -f "%Su" /usr/local/bin/brew
stat: cannot read file system information for ‘%Su’: No such file or directory
  File: "/usr/local/bin/brew"
    ID: 100000400000011 Namelen: *       Type: hfs
Block size: 4096       Fundamental block size: 4096
Blocks: Total: 60987521   Free: 18823880   Available: 18759880
Inodes: Total: 60987519   Free: 18759880
~  » stat -c "%U" /usr/local/bin/brew
rando
```

Before I start, is there any interest in a PR to fix this? Here's a fix to the same issue in another project: https://github.com/rtomayko/shocco/pull/17

PR #284, opened 2016-11-11
Prefer brew as package manager

DoomHammer:
[Linuxbrew](http://linuxbrew.sh/) is a fork of Homebrew aiming at Linux. As it is installable in users' home directory I think it should be prefered over system packagers.

`stat` workaround courtesy of #63

### DIFF
--- a/share/ruby-install/util.sh
+++ b/share/ruby-install/util.sh
@@ -3,11 +3,11 @@
 #
 # Auto-detect the package manager.
 #
-if   command -v apt-get >/dev/null; then package_manager="apt"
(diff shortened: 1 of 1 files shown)


FranklinYu:
How about allowing user to override the detection?

havenwood:
There's now the `--package-manager brew` option to explicitly select Homebrew as the package manager.

postmodern:
Preferring the use of `--package-manager brew`. We could possibly add additional logic to prefer homebrew, if it's installed and ruby-install detects we are running on macOS?

PR #290, opened 2017-01-06
added --download-only and --no-install options

LelandSindt:


LelandSindt:
@postmodern, can I get a little help with the failed tests?

postmodern:
What is the rational for these new options? Would you prefer a separate standalone util for fetching/extracting ruby archives?

LelandSindt:
I needed a way to programmatically pull and verify the latest ruby source, but not always install it. Contributing to this project seemed more useful what re-inventing the wheel. But yes, its not lost on me that having --no-install and --download-only options for a script designed to download and install might seem out of place. 

I did not write the tests because... I didn't realize you had automated testing... and... I have not had time to learn to write them.

Hopefully I have convinced you of the value... Perhaps you could point me to some "writing test scripts for dummies" resources.

postmodern:
I already have some code in [ruby-versions](https://github.com/postmodern/ruby-versions/blob/master/update.sh) which automatically downloads the archives for a version. Perhaps it could be extracted into a separate library/tool? Is your use-case general or specific?

The automated tests are [shunit2](http://ssb.stsci.edu/testing/shunit2/shunit2.html) tests. The individual tests are simply functions beginning with `test_` that call one or more of the `assert*` functions. The tests are grouped into separate files ending in `_test.sh` and ran individually by the `test/runner` script.

Issue #294, opened 2017-03-08
Adding a debug flag


eloyesp:
I would be an interesting adition and the code is not much, just `set -x` and that is enough.

postmodern:
This will satisfy #374 and #415.

postmodern:
The 0.9.0 branch now has a `-D,--debug` option that will output the relevant commands used to compile/install the ruby. fb4779c

PR #295, opened 2017-03-08
Add a debug flag to know what is happening

eloyesp:
As mentioned in #294 it just adds a debug flag to see what commands are being run.

I don't know how to add tests for it :(


havenwood:
On the testing front, it doesn't address the feature functionality, but it'd be nice to test the option parsing in `test/ruby-install-tests/parse_options_test.sh`.

eloyesp:
@havenwood I did that but it adds a xtrace for shunit with a lot of noise within the tests :)

havenwood:
@eloyesp Ahh, good point.

postmodern:
ruby-install 0.9.0 added a `-D,--debug` flag which prints out key commands (but not every command) that are ran during the installation process. Is `set -x` still needed?

PR #302, opened 2017-08-09
requires sudo or root

donbright:
the instructions in the README, when followed, run a script that runs 'sudo' to install dependencies. this doesnt work if you dont have sudo or root access.


havenwood:
You can skip dependencies with the `--no-install-deps` flag.

PR #303, opened 2017-09-27
Correct sudo comment

go2null:


go2null:
It seems the Tracis CI build is failing the setup stage, how do we resolve this?

postmodern:
Updated the tests to use a recent stable version (2.4) in the tests and rebased your doc fix.

PR #311, opened 2018-04-03
Use default-jdk virtual package for apt

ericmathison:
As per #276, Ubuntu 16.04 (Xenial) needs a newer jdk version as the openjdk-7-jdk package is no longer supported. Using the default-jdk package provides this and is available in the following Ubuntu versions:

* trusty (14.04LTS)
* xenial (16.04LTS)
* artful
* bionic

reference: https://packages.ubuntu.com/default-jdk

### DIFF
--- a/share/ruby-install/jruby/dependencies.txt
+++ b/share/ruby-install/jruby/dependencies.txt
@@ -1,4 +1,4 @@
-apt: openjdk-7-jdk
+apt: default-jdk
 dnf: java-openjdk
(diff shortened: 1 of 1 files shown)


havenwood:
@headius says 8 is recommended over default-jdk at the moment, since 9 and 10 still have known issues.

havenwood:
I love the idea of not hardcoding to a version, but it looks like #276 is best for now.

PR #314, opened 2018-06-15
Add TruffleRuby to ruby-install

eregon:
Hello,

Here is a PR to add [TruffleRuby](https://github.com/oracle/truffleruby) to ruby-install.
Corresponding PR for ruby-versions at https://github.com/postmodern/ruby-versions/pull/27.
I tried to keep the logic as simple as possible, and to update all relevant places mentioning other implementations.

I tested it locally and it seems to work well.

This is a fully open-source version of TruffleRuby, corresponding to a subset of the Community Edition of GraalVM (See http://www.graalvm.org/downloads/)
It includes the [`native` configuration](https://github.com/oracle/truffleruby#truffleruby-configurations) only, not the `JVM` configuration, so there is no JVM involved.

cc @havenwood

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -4,3 +4,4 @@ sudo make install
 ruby-install ruby
 ruby-install jruby
 ruby-install rubinius
+ruby-install truffleruby
--- a/share/ruby-install/truffleruby/functions.sh
+++ b/share/ruby-install/truffleruby/functions.sh
@@ -1,0 +1,37 @@
+#!/usr/bin/env bash
+
+# Check platform
+case $(uname) in
+	Linux) platform=linux ;;
+	Darwin) platform=macos ;;
+	*) fail "unknown platform $(uname)" ;;
+esac
+
+# Check architecture
+case $(uname -m) in
+	x86_64) arch=amd64 ;;
+	*) fail "unknown architecture $(uname -m)" ;;
+esac
+
+ruby_dir_name="truffleruby-$ruby_version-$platform-$arch"
+ruby_archive="$ruby_dir_name.tar.gz"
+ruby_mirror="${ruby_mirror:-https://github.com/oracle/truffleruby/releases/download}"
+ruby_url="${ruby_url:-$ruby_mirror/vm-$ruby_version/$ruby_archive}"
+
+#
+# Install TruffleRuby into $install_dir.
+#
+function install_ruby()
+{
+	log "Installing truffleruby $ruby_version ..."
+	cp -R "$src_dir/$ruby_dir_name" "$install_dir" || return $?
+}
+
+#
+# Post-install tasks.
+#
+function post_install()
+{
+	log "Running truffleruby post-install hook ..."
+	"$install_dir/lib/truffle/post_install_hook.sh" || return $?
+}
--- a/share/ruby-install/truffleruby/functions.sh
+++ b/share/ruby-install/truffleruby/functions.sh
@@ -1,17 +1,7 @@
 #!/usr/bin/env bash
 
-# Check platform
-case $(uname) in
-	Linux) platform=linux ;;
-	Darwin) platform=macos ;;
-	*) fail "unknown platform $(uname)" ;;
-esac
-
-# Check architecture
-case $(uname -m) in
-	x86_64) arch=amd64 ;;
-	*) fail "unknown architecture $(uname -m)" ;;
-esac
+platform=$(platform) || return $?
+arch=$(architecture) || return $?
 
 ruby_dir_name="truffleruby-$ruby_version-$platform-$arch"
 ruby_archive="$ruby_dir_name.tar.gz"
--- a/share/ruby-install/util.sh
+++ b/share/ruby-install/util.sh
@@ -121,3 +121,28 @@ function extract()
 			;;
 	esac
 }
+
+#
+# Returns the executing platform.
+#
+function platform()
+{
+	local uname="$(uname)"
+	case "$uname" in
+		Linux) echo linux ;;
+		Darwin) echo macos ;;
+		*) fail "unknown platform $uname" ;;
+	esac
+}
+
+#
+# Returns the CPU architecture.
+#
+function architecture()
+{
+	local arch="$(uname -m)"
+	case "$arch" in
+		x86_64) echo amd64 ;;
+		*) fail "unknown architecture $arch" ;;
+	esac
+}
--- a/share/ruby-install/util.sh
+++ b/share/ruby-install/util.sh
@@ -131,7 +131,10 @@ function platform()
 	case "$uname" in
 		Linux) echo linux ;;
 		Darwin) echo macos ;;
-		*) fail "unknown platform $uname" ;;
+		*)
+			error "unknown platform $uname"
+			return 1
+			;;
 	esac
 }
 
@@ -143,6 +146,9 @@ function architecture()
 	local arch="$(uname -m)"
 	case "$arch" in
 		x86_64) echo amd64 ;;
-		*) fail "unknown architecture $arch" ;;
+		*)
+			error "unknown architecture $arch"
+			return 1
+			;;
 	esac
 }
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Build Status](https://travis-ci.org/postmodern/ruby-install.svg?branch=master)](https://travis-ci.org/postmodern/ruby-install)
 
-Installs [Ruby], [JRuby], [Rubinius] or [mruby].
+Installs [Ruby], [JRuby], [Rubinius], [TruffleRuby] or [mruby].
 
 ## Features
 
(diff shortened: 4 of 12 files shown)


setup.sh
```diff
@@ -4,3 +4,4 @@ sudo make install
 ruby-install ruby
 ruby-install jruby
 ruby-install rubinius
+ruby-install truffleruby
```
eregon: I was not sure if I should add it here.

eregon:
I'll rebase this against the `0.7.0` branch so it can be in the next release hopefully, as @havenwood suggested.

share/ruby-install/truffleruby/functions.sh
```diff
@@ -0,0 +1,37 @@
...
+	Linux) platform=linux ;;
+	Darwin) platform=macos ;;
+	*) fail "unknown platform $(uname)" ;;
+esac
+
+# Check architecture
+case $(uname -m) in
+	x86_64) arch=amd64 ;;
+	*) fail "unknown architecture $(uname -m)" ;;
+esac
```
postmodern: Should we extract these into `util.sh`?

eregon: As you prefer, both are fine for me.
The resulting values might not be suitable for all usages though.
For instance, I can imagine some usages could expect `darwin` instead of `macos` for macOS.

havenwood: Can we get away with using `OSTYPE` and `MACHTYPE` since it's bash?

```bash
case "$OSTYPE" in
	linux*) platform=linux ;;
	darwin*) platform=macos ;;
	*) fail "incompatible platform $OSTYPE" ;;
esac

case "$MACHTYPE" in
	x86_64) arch=amd64 ;;
	*) fail "incompatible architecture $MACHTYPE" ;;
esac
```

eregon: I guess so, but I think the output of `uname` is more standard and does not vary as much (e.g., `MACHTYPE` is `x86_64-redhat-linux-gnu` on Bash, `x86_64` on ZSH and `OSTYPE` is `linux-gnu`).
`uname` is also what TruffleRuby uses upstream so we know it works for all our supported platforms.

havenwood: @eregon That makes sense.

I wonder if it's premature to extract this to `util.sh`, even though that ultimately seems like a good idea. Maglev was using `platform` and `arch`, but with it gone there's nothing else currently using them - and as pointed out, the desired `macos` and `amd64` values are domain specific.

eregon:
@postmodern @havenwood  I extracted 2 functions for platform & architecture in `util.sh`, and it indeed looks cleaner now. Thank you for the suggestion.

share/ruby-install/truffleruby/functions.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+platform=$(platform) || return $?
+arch=$(architecture) || return $?
```
eregon: I'm not sure if `return` is appropriate here (at the top-level of a `source`-d file) but it seems to work fine.
Using `exit` seems to have the same effect for me.
Note that it is needed to `return` or `exit` here, otherwise the script would continue if the platform/architecture is not found (and ultimately fail on download or extraction as `platform` would be an empty string).

Here is the output when the platform is not found:
```
$ bin/ruby-install --no-download truffleruby
!!! unknown platform MyPlatform
zsh: exit 255 
```

setup.sh
havenwood: I think it does make sense to add TruffleRuby here, assuming the point is to install the Ruby platforms that are commonly benchmarked these days. This file wasn't updated when the `--latest` flag was added, so that would probably also make sense to add across the board.

That said, I wonder if it'd be better to deprecate the feature and remove this file. @postmodern, should `setup.sh` still be a thing? If yes, should `--latest` be added?

share/ruby-install/truffleruby/functions.sh
havenwood: The #bash channel on Freenode.net seem to be in favor of `return` here, since it's being sourced inside a function and percolates up to a top-level `exit`.

An aside, but they also pointed out that `$?` is the implicit default status when one is omitted from `exit` and `return`. As a separate issue it might be worth considering removing all the `$?`s to reduce clutter.

donv:
Is this ready to merge?

Should it use `gu install org.graalvm.ruby` to install Ruby?

havenwood:
> Should it use gu install org.graalvm.ruby to install Ruby?

This is using the `truffleruby-*-amd64.tar.gz` tarball with Ruby already installed. It's been mentioned that it's possible the `gu install` command might change (it used to be `gu install -c`), so this preinstalled TruffleRuby tarball seems best.

eregon:
@donv It's ready for my side.
This is a lightweight native-only distribution, see the description of the PR.
To run on JVM, it's currently required to install GraalVM and the Ruby component manually.

@postmodern Could you review this and merge the PR?

share/ruby-install/util.sh
```diff
@@ -121,3 +121,28 @@ function extract()
...
+#
+# Returns the executing platform.
+#
+function platform()
+{
+	local uname="$(uname)"
+	case "$uname" in
+		Linux) echo linux ;;
+		Darwin) echo macos ;;
+		*) fail "unknown platform $uname" ;;
```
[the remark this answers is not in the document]
eregon: Note that due to this function returning a string via `echo`, the `exit` anyway does not affect the caller (beyond returning a non-zero exit status), as the caller uses `platform=$(platform) || return $?`.
The new commit is fine too and achieves the same semantics, I checked locally.

donv:
@havenwood @postmodern Any plans for a 0.7.0 release?

postmodern:
I think we're blocked on https://github.com/oracle/truffleruby/issues/1389. Either we have to print an error message if users run `ruby-install [--system|--install-dir ...] truffle-ruby`, or we wait for upstream to come up with a solution.

eregon:
TruffleRuby won't fix that issue for the next release because it's past feature freeze.

I think it's fundamentally flawed to install two different Ruby in the same prefix: they will simplify override each's other `bin/ruby` (and also `bin/rake`, `bin/irb`, `bin/gem`, etc) so this seems to never work in practice.

Installing a single Ruby under `/usr/local` works, but triggers complications to uninstall it.
* For MRI, if one keeps the `~/src/ruby-X.Y.Z` directory, then `make uninstall` can be run to remove most files. But some files are left, in my case `bin/rake` and 7 gemspecs in `lib/ruby/gems/2.4.0/specifications`. So this still requires manually removing files when uninstalling.
* I don't see any way to uninstall JRuby. Maybe if `~/src/jruby-VERSION` files are kept one can use them as a list of files to remove.
* I don't see any way to uninstall Rubinius. Running `bundle exec rake` in `~/src/rubiniues-VERSION` is impractical as that will show `Your configuration is outdated, please run ./configure first` unless you're lucky to have the exact same Ruby that was used to build Rubinius (no idea how to figure that).

So while a single Ruby under `/usr/local` works, it seems problematic if it must be uninstalled.
I would not recommend installing to `/usr/local`, no matter the Ruby implementation because of this.

With the current TruffleRuby layout, I think it's not a problem if it's the only Ruby installed in `/usr/local`. The directory structure shown in https://github.com/oracle/truffleruby/issues/1389#issuecomment-404491193 seems unlikely to conflict with other software. It can be uninstalled by using the list of files under, e.g., `~/src/truffleruby-1.0.0-rc3-linux-amd64` similar to JRuby.
@postmodern  So I'm not sure we need a warning. But feel free to add one when installing to `/usr/local` if you think it's important.

eregon:
> Are some dangling top-level files enough of an eye sore to hold this back?

@postmodern It's not ideal but they probably don't hurt anything either. They are all additional information/documentation and not required to run TruffleRuby (but the licenses are required for distribution).
I'd rather wait to see if people actually care about that because I expect very few install to `/usr/local` and if they do they must already be aware this is complicated to uninstall and so probably only works well in virtualized environments (e.g., with Docker where one can just rebuild from the start).

> Wouldn't installing truffleruby alongside (or more accurately, over) MRI would just overwrite MRI's bin/ scripts, thus changing their #! to point to truffleruby?

This is extremely brittle, to install any Ruby over another Ruby (or version).
This will leave some files behind of the old installation, and that might very well break both the old and new Ruby. So I would definitely warn in such a case.
I would even prevent doing this as it's just wrong.
Requiring confirmation (with `--reinstall`) can work too, but I don't see a single situation where this is safe, except if it's exactly the same Ruby from the same archive (but then what's the point?).

eregon:
@postmodern FWIW, JRuby also has top-level files: `BSDL COPYING LEGAL LICENSE.RUBY`.
And it seems nobody complained so far, so I think this doesn't matter.

eregon:
@postmodern In any case, I believe none of this is really new or changed by TruffleRuby, it was already like that or similar with JRuby and ruby-install for many years. So I think there is no need to fix what you mentioned above for 0.7.0.

postmodern:
OK, we'll go forward with this. I'll try to find time this week (or weekend) to prepare a 0.7.0 release.

eregon:
Great :)

PR #318, opened 2018-07-16
Added 'bzip2' dependency for RedHat/Fedora/CentOS-based distributions

epitron:
I just ran into this issue on a Centos7 VM.


havenwood:
Rubinius also uses bzip2.

Right now ruby-install only installs Ruby dependencies, not its own deps like bash, wget or curl, tar and patch.

P.S. - Hi @epitron!

epitron:
Ah, I see -- trying to avoid package bloat. (Lord knows I hate it when I find a box that has tar or bash on it!)

But seriously, installing "gcc automake bison zlib-devel libyaml-devel openssl-devel gdbm-devel readline-devel ncurses-devel libffi-devel", then drawing the line at "bzip2" is quite silly. :)

tar/bzip2 are build dependencies, in that you can't build anything without it. I'm not sure what argument could be made in support of the idea that a machine should have gcc/bison/zlib-devel installed, but not bzip2.

&lt;/rant&gt;

PS: hi how's it goin! :D

postmodern:
We are in the process of switching ruby to downloading `.tar.xz` archives and adding `xz-utils` as a dependency. See the [use_tar_xz_for_ruby](https://github.com/postmodern/ruby-install/compare/0.8.0...use_tar_xz_for_ruby) branch.

PR #329, opened 2018-12-14
Transform ruby name to lowercase

papilip:
Hello,

Today, I wanted to use ruby-install by copying / pasting the new `Ruby 2.6.0-rc1` version from the blog. And your utility said `!!! Ruby Unknown: Ruby`

I think it's not a good idea to not allow capitalization. I send you this PR with the hope that you like it.

I wish you a Merry Christmas with your family


postmodern:
Eh, maybe it might be better to print the list of supported rubies if the ruby name is unknown?

Issue #328, opened 2018-12-12
Fix the help for the --jobs option


RobertAudi:
The help for the `--jobs` option is missing the `=` sign; The [man page](https://github.com/postmodern/ruby-install/blob/master/doc/man/ruby-install.1.md#options) documents the option properly.

https://github.com/postmodern/ruby-install/blob/adaeb23ff136270b3c8526554f1f17fe6fbc57e0/share/ruby-install/ruby-install.sh#L43

https://github.com/postmodern/ruby-install/blob/ea2b8bbff87ea4340d965e72ac3acfb4ab1d1766/share/man/man1/ruby-install.1#L35

The error message thrown when omitting the `=` is kind of misleading:

```shell
$ ruby-install --jobs 4
!!! Unknown ruby: 4
```

The reason is that the number of jobs is used as the desired ruby version, even if the `--jobs` option is not the last one:

```shell
$ ruby-install --jobs 4 --cleanup --no-reinstall ruby
!!! Unknown ruby: 4
```

postmodern:
Fixed in a0ee3814731061559852d775cbca178b6d76ae25 and 0a683f0e60a7804a7066b608a56f1c46c2e0cd93.

PR #330, opened 2019-01-03
Fix the usage help to respect the correct usage of the --jobs option

RobertAudi:
The `--jobs` options requires an `=` sign to work. Without the equal sign, the number of jobs is interpreted as the ruby version to install.

The problem is described in more details in #328

### DIFF
--- a/share/ruby-install/ruby-install.sh
+++ b/share/ruby-install/ruby-install.sh
@@ -40,7 +40,7 @@ Options:
 	    --system		Alias for -i $system_dir
 	-s, --src-dir DIR	Directory to download source-code into
 	-c, --cleanup		Remove archive and unpacked source-code after installation
-	-j, --jobs JOBS		Number of jobs to run in parallel when compiling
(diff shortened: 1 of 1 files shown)


simi:
:+1: it is a little confusing, this will make it clear

havenwood:
The other option would be to change the parsing so it doesn't require the equals. I thought the short and long form flags could just be passed through to make, but as noted the long form follows the equals convention.

If we leave it as is, it's a great idea to document. I'd lean toward fixing it instead, so `--jobs 8` works rather than `--jobs=8`.

cbandy:
Looks like this conflicts with https://github.com/postmodern/ruby-install/commit/0a683f0e60a7804a7066b608a56f1c46c2e0cd93 which has a similar fix.

postmodern:
Fixed in 0a683f0e60a7804a7066b608a56f1c46c2e0cd93 and 3e37a2efee060c5b5a8366118f40853a408de88d.

PR #332, opened 2019-01-24
add hyphen to syntax of command line ruby argument for correct version install

kamillamagna:
Running the arguments ```ruby-install ruby x.y.z``` as per current README installs the latest stable version of ruby instead of version x.y.z as it is missing a hyphen between the ruby and the version.
Changed all instances in README of ```ruby x.y.z``` to ```ruby-x.y.z```.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,11 +64,11 @@ Install the latest version of Ruby:
 
 Install a stable version of Ruby:
 
-    $ ruby-install ruby 2.3
+    $ ruby-install ruby-2.3
 
 Install a specific version of Ruby:
 
-    $ ruby-install ruby 2.2.4
+    $ ruby-install ruby-2.2.4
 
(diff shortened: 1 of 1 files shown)


havenwood:
I'm not able to reproduce this issue. I'm seeing the correct version selected with both variant of the syntax. Can you still reproduce this with the latest version of ruby-install?

An aside, but enough folk seem confused by the space-delimited variant that it does seem nice to me to show `ruby-x.y.z` in examples even though both work.

postmodern:
Originally `ruby-install` was designed to take two separate arguments (`ruby` and `x.y.z`), not a hyphenated `ruby-x.y.z` argument. Only just recently, support for specifying a hyphenated `ruby-x.y.z` or even just a single `x.y.z` argument was added in 7339df6e1f1775253db11c189ca41d6ebe72fb7a.

PR #333, opened 2019-01-29
Add missing grep and awk dependencies for pacman

alyssais:
The default Arch Docker image doesn't have these installed.

Reproduction:

```dockerfile
FROM archlinux/base

RUN pacman -Syu --noconfirm sudo make && \
    curl -Lo ruby-install-master.tar.gz https://github.com/postmodern/ruby-install/archive/master.tar.gz && \
    tar -xzvf ruby-install-master.tar.gz && \
    cd ruby-install-master && \
    make install && \
    pacman -Syu --noconfirm grep awk && \
    yes | ruby-install ruby
```

### DIFF
--- a/share/ruby-install/ruby/dependencies.txt
+++ b/share/ruby-install/ruby/dependencies.txt
@@ -3,6 +3,6 @@ dnf: gcc automake bison zlib-devel libyaml-devel openssl-devel gdbm-devel readli
 yum: gcc automake bison zlib-devel libyaml-devel openssl-devel gdbm-devel readline-devel ncurses-devel libffi-devel
 port: automake bison openssl readline libyaml gdbm libffi
 brew: automake bison openssl readline libyaml gdbm libffi
-pacman: gcc make bison zlib ncurses openssl readline libyaml gdbm libffi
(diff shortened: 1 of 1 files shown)


postmodern:
@alyssais should these dependencies be added to any other rubies that compile?

postmodern:
@alyssais also, is this an issue that should be fixed in the Arch Docker image, or is it intentionally bare bones?

alyssais:
> @alyssais should these dependencies be added to any other rubies that compile?

No other ruby that compiles seems to need these.  Rubinius seems like it
would, but doesn't compile anyway, I think because Arch's clang is too
new (and it doesn't offer older versions).

> @alyssais also, is this an issue that should be fixed in the Arch
> Docker image, or is it intentionally bare bones?

I assume it must be intentional -- it would be very difficult for it to
be accidental because of how obvious it is, since almost nothing works. ;)

postmodern:
We could add `awk`, but `ruby-install` uses `grep` internally before it even gets to installing dependencies, so that wouldn't work.

Also, `ruby-install` is not necessarily needed to install a ruby into a docker image. It's probably easier to add the commands to install the dependencies, download the `.tar.xz` archive, extract, `configure --prefix=... && make && make install`.

PR #343, opened 2019-04-01
Don't sudo if you are the brew owner

kwstannard:
Probably a better way to do this, but I just needed to install ruby on a system that I don't have sudo access.

### DIFF
--- a/share/ruby-install/package_manager.sh
+++ b/share/ruby-install/package_manager.sh
@@ -3,6 +3,7 @@
 #
 # Auto-detect the package manager.
 #
+
 function detect_package_manager()
 {
 	if   command -v zypper  >/dev/null; then package_manager="zypper"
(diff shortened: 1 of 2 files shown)


postmodern:
Could we do the brew ownership check at the top of `package_manager.sh`, and have a function called `brew_sudo` or something that checks whether we need sudo or not? This would prevent having to duplicate the two `brew` commands.

share/ruby-install/package_manager.sh
```diff
@@ -3,6 +3,16 @@
 #
 # Auto-detect the package manager.
 #
+
+brew_owner="$(/usr/bin/stat -f %Su "$(command -v brew)")";
```
postmodern: Style: commands do not need to end with a semicolon.

share/ruby-install/package_manager.sh
```diff
@@ -3,6 +3,16 @@
...
+brew2="";
+
+if [ "$(id -un)" == "$brew_owner" ]; then
+  brew2="brew"
```
postmodern: Style: shell script typically use hard tab indentation.

share/ruby-install/package_manager.sh
```diff
@@ -3,6 +3,16 @@
 #
 # Auto-detect the package manager.
 #
+
+brew_owner="$(/usr/bin/stat -f %Su "$(command -v brew)")";
+brew2="";
```
postmodern: Could use a slightly better variable name. Maybe `brew_cmd`?

share/ruby-install/package_manager.sh
postmodern: Also, this code should probably go inside of the `detect_package_manager` function, under the `elif command -v brew` branch. No point in running this code if brew isn't installed.

kwstannard:
I was also trying to figure out if there is a reason this couldn't apply to all pkg managers. I didn't try it here because I don't have time to check if it would work.

postmodern:
To my knowledge, most system package managers that come pre-installed (apt, dnf, etc) require root. Brew is unique since it can be installed by the user.

kwstannard:
You are correct. I was thinking that since comparing owner to the user would essentially replicate the $sudo functionality we could replace it and have all paths go through the same code. Unfortunately I found out that GNU stat has an extremely different implementation to BSD stat.

share/ruby-install/package_manager.sh
```diff
@@ -32,30 +33,29 @@ function set_package_manager()
...
-		"")	warn "Could not determine Package Manager. Proceeding anyway." ;;
+		"")	    warn "Could not determine Package Manager. Proceeding anyway." ;;
+		brew)   installer install "$@" || installer upgrade "$@" || return $? ;;
+		zypper) installer -n in -l $* || return $? ;;
+		*)      installer install -y "$@" || return $?;;
 	esac
 }
 
+function installer() {
+	p="$(command -v $package_manager)"
```
postmodern: You need the `local` keyword to prevent leaking variables.

share/ruby-install/package_manager.sh
```diff
@@ -32,30 +33,29 @@ function set_package_manager()
...
 
 			if (( ${#missing_pkgs[@]} > 0 )); then
-				$sudo pacman -S "${missing_pkgs[@]}" || return $?
+				installer -S "${missing_pkgs[@]}" || return $?
 			fi
 			;;
-		zypper) $sudo zypper -n in -l $* || return $? ;;
-		"")	warn "Could not determine Package Manager. Proceeding anyway." ;;
+		"")	    warn "Could not determine Package Manager. Proceeding anyway." ;;
+		brew)   installer install "$@" || installer upgrade "$@" || return $? ;;
```
postmodern: It doesn't make sense that we're checking `$package_manager` but have a generic `installer` function; which is kind of what `install_packages` should be. We should either define different `function install_packages()` based on the `if/elif command -v ...` checks:

    if command -v dnf >/dev/null; then
          function install_packages()
          {
              $sudo dnf install "$@"
          }

or auto-detect everything in `detect_package_manager` (ex: `$sudo`) which `install_packages` would then use. Also note that `$sudo` can be defined as `sudo -u "$brew_owner"`.

share/ruby-install/package_manager.sh
kwstannard: Good catch.

share/ruby-install/package_manager.sh
FranklinYu: `port -y`?

goofrider:
Upvote +1. 

Why not just merge the original pull request if it works? I really don't want to grant my test environments and users sudo permission.

postmodern:
@goofrider I can try rewriting this. All of the checks can be done at load-time, at the top of the file where we detect the package manager. Also no reason to store the commands in a `.txt` file since they won't be changed often.

kwstannard:
@postmodern I am good with a rewrite.

kwstannard:
I am no longer working at a place that requires this patch. Going to close this so that it isn't clogging up my PR index.

PR #348, opened 2019-07-07
fix(uninstall): DO NOT ERASE /usr/local/share

yumetodo:
bacause of b87e5955f770491ede60e41404badd91a6610ccc
make uninstall fail and produce error below:

```
for file in `find bin share -type f`; do rm -f /usr/local/$file; done
rm -rf /usr/local/share/doc/ruby-install-0.7.0
rmdir /usr/local/share
rmdir: '/usr/local/share': Directory not empty
make: *** [Makefile:78: uninstall] Error 1
```

We should just remove `/usr/local/share/ruby-install`, not `/usr/local/share`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,6 @@ install:
 uninstall:
 	for file in $(INSTALL_FILES); do rm -f $(DESTDIR)$(PREFIX)/$$file; done
 	rm -rf $(DESTDIR)$(DOC_DIR)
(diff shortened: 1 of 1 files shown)


yumetodo:
any updates?

postmodern:
@yumetodo think this was fixed in 07a3694825d980d520f1433765004e0390ef8933. Thank you for reporting!

PR #354, opened 2019-09-28
Use sudo -E to pass environment variables through to brew.

zenspider:
`brew` uses environment variables to control many aspects of how it
operates. Using `sudo` with `-E` allows those variables to be used to
keep the build/setup consistent with the existing environment.

### DIFF
--- a/share/ruby-install/package_manager.sh
+++ b/share/ruby-install/package_manager.sh
@@ -25,7 +25,7 @@ function install_packages()
 			local brew_sudo=""
 
 			if [[ "$brew_owner" != "$(id -un)" ]]; then
-				brew_sudo="sudo -u $brew_owner"
+				brew_sudo="sudo -Eu $brew_owner"
 			fi
 
 			$brew_sudo brew install "$@" ||


zenspider:
poke?

postmodern:
@zenspider are there specific env variables that need to be passed thru? Preserving the whole environment is a potential security risk. `sudo --preserve-env=...` would be safer.

zenspider:
For starters:

```
2532 % env | grep BREW
HOMEBREW_AUTO_UPDATE_SECS=86400000
BREW_CASK_OPTS=--appdir=~/Applications
HOMEBREW_NO_SANDBOX=1
HOMEBREW_NO_ANALYTICS=1
HOMEBREW_NO_AUTO_UPDATE=1
HOMEBREW_NO_GITHUB_API=1
HOMEBREW_NO_INSTALL_CLEANUP=1
HOMEBREW_NO_UPDATE_CLEANUP=1
```

and whatever other annoying thing they add in the future... 🤷‍♀️

The security policy can determine what is risky and what is not (at least in a black and white sense).

havenwood:
Maybe pass all env vars beginning with `HOMEBREW_`, since there are many and new ones like @zenspider points out?

zenspider:
Annual poke...

dustin-cowles:
Please! It is super annoying when installing a new ruby version kicks off a 30 minute long brew upgrade of unrelated packages.

postmodern:
OK ok, I like @havenwood's idea of only allowing certain `HOMEBREW_*` env variables. We could use `--preserve-env=list` instead of `-E` to only pass in known `HOMEBREW_` environment variables and avoid any potential issues with the user's entire env unintentionally leaking something into homebrew's env.

postmodern:
It just occurred to me that using `sudo -E` or even `sudo --preserve-env=HOMEBREW_*,...` would not work _if_ you invoked `ruby-install` with `sudo` initially. The initial `sudo` would filter out any `HOMEBREW_*` variables, before `sudo -E`/`sudo --preserve-env=...` could pass them to `brew install`. Perhaps a better solution might be to [configure `/etc/sudoers`](https://man7.org/linux/man-pages/man5/sudoers.5.html) to explicitly allow the `HOMEBREW_*` env variables?

    Defaults env_keep += "HOMEBREW_NO_SANDBOX"
    Defaults env_keep += "HOMEBREW_NO_ANALYTICS"
    ...

postmodern:
@monfresh since `ruby-install` can be ran as a regular user *or* as root via `sudo`, we have to run homebrew under the user which installed homebrew. We should probably not be running homebrew as root. The logic in [`install_packages`](https://github.com/postmodern/ruby-install/blob/4971a1f934a687c1032e5e17615f791743e8c5d4/share/ruby-install/package_manager.sh#L32-L40) need to be refined. I will create a new issue specifically about that.

@zenspider I am very hesitant about merging anything which might introduce potential security issues, such as passing *every* environment variable from one user's session to another command (there's a reason why `sudo` does not preserve the caller's environment by default). I mentioned various problems with `sudo -E` and some alternative approaches, but nothing has changed with this PR. I will close this PR and open a new issue.

postmodern:
See issue #471. @monfresh I suspect a better solution specifically for homebrew being erroneously ran as `root` would be running the `brew` commands as the *original* non-root user (aka `$SUDO_USER`).

PR #356, opened 2019-10-08
add CentOS7/8 support

s3fxn:
patch for handling distro specific dependencies on CentOS 7/8


postmodern:
I think we'll need a more extendable solution for distro/release specific package names. We've ran into this in the past with Ubuntu and Debian. The simple solution I've tried to follow is only support the packages from the most recent stable release.

postmodern:
I'm curious how RedHat's shift to CentOS Stream might change this?

PR #373, opened 2020-07-12
Add support for installing GraalVM

eregon:
* Supports using other GraalVM languages such as JavaScript, Python, R and Java.
* Supports running TruffleRuby on JVM and not only Native (see https://github.com/oracle/truffleruby#truffleruby-runtime-configurations).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,10 @@ clean:
 check:
 	shellcheck --exclude SC2034 share/$(NAME)/*.sh bin/*
 
+lint:
+	# Check dependencies are consistent for truffleruby and graalvm
+	diff share/ruby-install/graalvm/dependencies.txt share/ruby-install/truffleruby/dependencies.txt
+
 test:
 	./test/runner
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,4 +21,6 @@ before_script:
   - __rvm_unload
   - export SHUNIT2="$(command -v shunit2 2>/dev/null)"
 
-script: make test
+script:
+  - make lint
+  - make test
--- a/share/ruby-install/graalvm/dependencies.txt
+++ b/share/ruby-install/graalvm/dependencies.txt
@@ -1,0 +1,8 @@
+apt: zlib1g-dev libssl-dev make gcc libxml2
+dnf: zlib-devel openssl-devel make gcc libxml2
+yum: zlib-devel openssl-devel make gcc libxml2
+port: openssl
+brew: openssl
+pacman: zlib openssl make gcc libxml2
+zypper: zlib-devel libopenssl-devel make gcc libxml2
+pkg: openssl gmake gcc libxml2
--- a/share/ruby-install/graalvm/functions.sh
+++ b/share/ruby-install/graalvm/functions.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+platform=$(platform) || return $?
+graalvm_platform="${platform/macos/darwin}"
+arch=$(architecture) || return $?
+
+ruby_dir_name="graalvm-ce-java8-$ruby_version"
+ruby_archive="graalvm-ce-java8-$graalvm_platform-$arch-$ruby_version.tar.gz"
+ruby_mirror="${ruby_mirror:-https://github.com/graalvm/graalvm-ce-builds/releases/download}"
+ruby_url="${ruby_url:-$ruby_mirror/vm-$ruby_version/$ruby_archive}"
+
+#
+# Install GraalVM into $install_dir.
+#
+function install_ruby()
+{
+	log "Installing GraalVM $ruby_version ..."
+	cp -R "$src_dir/$ruby_dir_name" "$install_dir" || return $?
+}
+
+#
+# Post-install tasks.
+#
+function post_install()
+{
+	log "Installing the Ruby component ..."
+	"$install_dir/bin/gu" install ruby || return $?
+
+	log "Running truffleruby post-install hook ..."
+	local ruby_home
+	ruby_home=$("$install_dir/bin/ruby" -e 'print RbConfig::CONFIG["prefix"]')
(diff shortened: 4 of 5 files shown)


eregon:
CI is green on Linux, and I tested locally it installs fine.
On macOS, there seem to be issues with shunit2: https://travis-ci.org/github/postmodern/ruby-install/jobs/707341527
IMHO we shouldn't prevent merging this PR for that since it didn't cause it.

eregon:
@postmodern @havenwood Could you review?

Makefile
```diff
@@ -53,6 +53,8 @@ check:
 
 test:
 	./test/runner
+	# Check dependencies are consistent for truffleruby and graalvm
+	diff share/ruby-install/graalvm/dependencies.txt share/ruby-install/truffleruby/dependencies.txt
```
postmodern, line 57: I think we should add another task called `lint` to do various sanity/QA checks.

eregon: Sounds good, will do.

eregon: Done

postmodern:
Manually merged into the new [0.8.0](https://github.com/postmodern/ruby-install/tree/0.8.0) branch.

eregon:
Thanks!

Do you plan to release 0.8.0 soon?

It would be helpful to both be able to install GraalVM with ruby-install, and also to get the fix from #359 (https://github.com/postmodern/ruby-install/pull/359#issuecomment-627431864).

eregon:
@postmodern note we'll need https://github.com/postmodern/ruby-versions/pull/49 too to be able to list versions and to verify checksums.

eregon:
Note: I might want to refine the setup to only have Ruby executables under `bin/` and not all GraalVM executables.
I'll try to to make a PR to do that soon.

postmodern:
@eregon if we could install graalvm with only the `bin/ruby` executable, i'd also prefer renaming it to `graalvm-ruby`.

eregon:
Do you mean with only the Ruby executables like done [in ruby-build](https://github.com/rbenv/ruby-build/pull/1455#issuecomment-657753289)?

For the name, should we use `truffleruby+graalvm` too (or `truffleruby-graalvm` if that works)?
`graalvm-ruby` seems rather confusing to me and the relation to `truffleruby` is then not clear.

postmodern:
Hmm, then probably `graalvm-truffleruby`, since graalvm is downloaded/unpacked then the `lib/truffle/post_install_hook.sh` script is ran. `truffleruby-graalvm` would also be acceptable.

I would prefer only having the ruby executable(s) installed. Would seem like scope-creep if `ruby-install` started to install JavaScript, Python, R and Java executables.

eregon:
Using `-` in the name seems problemtic for ruby-install too:
```
bin/ruby-install truffleruby-graalvm
*** Unknown truffleruby version graalvm. Proceeding anyways ...
>>> Installing truffleruby graalvm into /home/eregon/.rubies/truffleruby-graalvm ...
>>> Installing dependencies for truffleruby graalvm ...
```
vs the normal output:
```
bin/ruby-install truffleruby        
>>> Installing truffleruby 20.2.0 into /home/eregon/.rubies/truffleruby-20.2.0 ...
>>> Installing dependencies for truffleruby 20.2.0 ...
```

eregon:
@postmodern I implemented those changes in https://github.com/postmodern/ruby-install/pull/377

postmodern:
@eregon looks like `parse_ruby` is naively interpreting `truffleruby-graalvm` as being of the form `ruby-version`. We could probably improve the matching `case` regexp to look for versiony looking strings. Probably something as simply as `*-[0-9]*` might work.

eregon:
@postmodern Would `truffleruby+graalvm` be OK or you want `truffleruby-graalvm`?
IIRC for ruby-build it was basically impossible to parse correctly if there was `-` in the name (notably for sorting versions and filtering stable versions, I remember trying to fix it). Not sure for `ruby-install`.

postmodern:
@eregon it might be possible with a `[[ ... =~ "..." ]]` regexp match and using `BASH_REMATCH` (bash >= 3) or `MATCH` (zsh). Looks like ruby-build already uses `truffleruby+graalvm`, we could go with that out of respect for consistency.

postmodern:
I noticed that the `darwin` archive has everything within `Contents/Home`. Would it be possible to copy everything out of `Contents/Home/*` into `$install_dir/graalvm`, or would that cause issues?

eregon:
`Contents/Home` is the standard layout for all JDKs on macOS, and this is how GraalVM is officially distributed, so it should be kept.
Does it cause any issue?

postmodern:
Ideally, I'd like truffleruby+graalvm to be in FHS structure. Not sure if that's possible though? Using the FHS directory structure would allow it to cleanly install into `/usr/local/`, as well as `/opt/`. It also helps distro package maintainers to package truffleruby+graalvm.

eregon:
There is https://github.com/oracle/truffleruby/issues/1389 for FHS structure. It's not really a priority at the moment.
In my experience, Ruby provided by package managers is often a nightmare to use (for gems, etc), so that doesn't seem interesting right now.
I'd be OK to add a check that truffleruby/truffleruby+graalvm are installed under `$HOME` (and error otherwise), if you want to add one.
I think it's quite rare that people install ruby globally, unless they use a package manager.

Makefile
postmodern: @eregon random question, could the dependencies of truffleruby vs truffleruby-graalvm ever differ? If they absolutely must stay locked, I am considering moving the diff into it's own shunit2 test. Also, we could symlink `truffleruby-graalvm/dependencies.txt` back to `truffleruby/dependencies.txt` to ensure they are always the same.

eregon: I guess they might differ at some point, if it became necessary.
This check is intended to remind us to check dependencies of one when changing the other.

PR #377, opened 2020-09-19
Rename to truffleruby+graalvm, only add Ruby executables in PATH, clean installation directory before installing

eregon:
Addresses https://github.com/postmodern/ruby-install/pull/373#issuecomment-662414406 and the following comments.


eregon:
This needs a corresponding change in `ruby-versions` which I have here: https://github.com/eregon/ruby-versions/commit/f40e0c032744da6fee7739359964603af880979b
I'll push that to `ruby-versions` `master` once this is merged.
We don't need to care about the older name (`graalvm`) since `ruby-install` has no release with it.

postmodern:
@eregon could you go ahead and push your changes to ruby-versions, then switch the `ruby_versions_url` back to the official one?

eregon:
@postmodern Done, and PR updated.

postmodern (changes requested):
This should be removed. It would cause massive problems if the installation directory is something like `/usr/local/`. Ideally, the ruby's installation process should cleanly overwrite all files.

eregon:
@postmodern Do you mean 46d2d2c? I can remove that from this PR.

> Ideally, the ruby's installation process should cleanly overwrite all files.

I don't think that's realistic. Any leftover file can cause unpredictable failures that nobody wants to investigate.
Simple example: a stdlib file (or a bin/foo file) is removed, it would still be there if there was a previous installation and there is no such cleanup. And then you end up with "it works on my machine but not on yours" situations.

I see your point that it's problematic for `/usr/local`. Maybe we can detect if `ruby-install` installs to a directory with the ruby name in it, in which case it should be fairly safe? (e.g. removing `~/.rubies/ruby-2.6.5` when re-installing Ruby 2.6.5 makes perfect sense)

eregon:
@postmodern I removed that commit from this PR.

eregon:
Thanks for merging!

postmodern:
FYI I have merged the [better_parse_ruby](https://github.com/postmodern/ruby-install/commit/7339df6e1f1775253db11c189ca41d6ebe72fb7a) branch into 0.8.0. This should now be able to handle `ruby`, `version` and `ruby-version` arguments. Unless there's an important reason why truffleruby+graalvm should contain the `+`, I would like to rename it to either truffleruby-graalvm or graalvm-truffleruby.

eregon:
@postmodern Sounds nice, thanks.
I think `truffleruby-graalvm` is good.
I don't like `graalvm-truffleruby` because that sounds to me like all graalvm executables should be available.

donv:
Regular truffleruby will still be available, right?

eregon:
@donv Yes, `truffleruby` is TruffleRuby Native, and `truffleruby-graalvm` is TruffleRuby Native + JVM + ability to install and interop with other languages.

postmodern:
@eregon what URL should I use for truffleruby-graalvm and how should I refer to it in the README? "TruffleRuby (GraalVM)"?

eregon:
`TruffleRuby (GraalVM)` sounds good.

I will add instructions how to install truffleruby-graalvm in https://github.com/oracle/truffleruby/blob/master/doc/user/ruby-managers.md with `ruby-install` when it's released.
Linking there doesn't seem very useful though, so I'd still link to the README: https://github.com/oracle/truffleruby

postmodern:
@eregon if both TruffleRuby's will link to the same place, could I combine them into something like "TruffleRuby (native / GraalVM)"?

eregon:
Yeah, that sounds good :)

Issue #128, opened 2014-02-24
proxy support


epinault:
On OSX Maverick (10.9.1), using bash shell  3.2.51(1), and ruby-install 0.4.0

I love ruby-install but does not seem to work with proxy.

I have my http_proxy and https_proxy correctly and can do a curl manually and get the files.. but does not seem to work when running  "ruby-install ruby 2.1.1"

I get bunch of timeout. So I suspect those ENV variable are not passed/defined in the shell executed by ruby-install

postmodern:
If anyone wants to submit a PR to pass proxy options to both `curl` and `wget`, I'd accept it!

skade:
I hit this issue when ruby-install tries to install via brew. Calling brew directly works.

sofaking:
As an addition to @skade comment, I tried to preserve proxy settings via sudoers file (```/private/etc/sudoers```). You'll need to add your settings to ```env_keep``` section.

Issue #376, opened 2020-09-18
Installing with HomeBrew 2.5.1 broken


chrt00:
Looks like you can no longer run `brew` with sudo.

```
$ ruby-install ruby-2.4
>>> Installing ruby 2.4.10 into /Users/---/.rubies/ruby-2.4.10 ...
>>> Installing dependencies for ruby 2.4.10 ...
Error: Running Homebrew as root is extremely dangerous and no longer supported.
As Homebrew does not drop privileges on installation you would be giving all
build scripts full access to your system.
Error: Running Homebrew as root is extremely dangerous and no longer supported.
As Homebrew does not drop privileges on installation you would be giving all
build scripts full access to your system.
!!! Installing dependencies failed!

```

postmodern:
Could you specify the version of ruby-install you are using and the output of this command `/usr/bin/stat -f %Su "$(command -v brew)"`.

rb83:
@postmodern see PR #396.

postmodern:
#396 has been merged which should fix the homebrew errors. Will release 0.8.2 sometime this week.

postmodern:
0.8.2 has been tagged. Now we just have to wait for homebrew to pull in the new formula.

rb83:
Cool, thanks for your work @postmodern

PR #396, opened 2021-02-05
Do not use sudo for brew if not necessary

rb83:
Normally, brew installations do not require sudo. There might be
certain cases where it does, so this commit preserves the old behaviour
where ownership of the brew script isn't the same as the user executing
ruby-install. Otherwise, sudo is omitted.

This also solves #128 for situations where the issue was caused by sudo
not preserving the user's environment (proxy failures could also be
caused by a host of other reasons, but it's the main reason why the issue was created originally).


rb83:
CI failures seem to have been caused in unrelated files?

rb83:
rebased on upstream master in order to have the checks re-run.

rb83:
Also solves #376.

postmodern:
0.8.2 has been tagged. Now to wait for homebrew to pull in the new formula.

Issue #398, opened 2021-04-07
Track TruffleRuby FHS issue


missingcharacter:
The issue I'm talking about is https://github.com/oracle/truffleruby/issues/1389 and the reason I think this issue should be linked to this project is `truffleruby` cannot be currently installed using `--system` (`/usr/local`), even when the issue is not something this project will fix users of `ruby-install` are affected by it and seeing a connection between the 2 projects would probably save people time when they realize `truffleruby` cannot be installed using `--system`.

Note: I only became aware of this problem after I noticed `truffleruby` was supported, see below:

```shell
ruby-install --latest
>>> Downloading latest ruby versions ...
>>> Downloading latest jruby versions ...
>>> Downloading latest rbx versions ...
>>> Downloading latest truffleruby versions ...
>>> Downloading latest truffleruby-graalvm versions ...
>>> Downloading latest mruby versions ...
Stable ruby versions:
  ruby:
    2.6.7
    2.7.3
    3.0.1
  jruby:
    9.2.16.0
  rbx:
    5.0
  truffleruby:
    21.0.0
  truffleruby-graalvm:
    21.0.0
  mruby:
    2.1.2
```

and later I tried to install `truffleruby` in a container

eregon:
Maybe `ruby-install` should warn or error when trying to install truffleruby with `--system`?

I wouldn't bet on https://github.com/oracle/truffleruby/issues/1389 being solved very soon, it's a large effort for what seems a fairly low gain, most rubyists already install Rubies in their own directory and not randomly mix them with system files (which feels like it always causes more problem than it solves).

missingcharacter:
@eregon and @postmodern what do you think of this? https://github.com/postmodern/ruby-install/pull/401

PR #401, opened 2021-05-29
Fail truffleruby installation if install_dir is /usr/local

missingcharacter:
Trying to close https://github.com/postmodern/ruby-install/issues/398

### DIFF
--- a/share/ruby-install/truffleruby-graalvm/functions.sh
+++ b/share/ruby-install/truffleruby-graalvm/functions.sh
@@ -21,6 +21,11 @@ ruby_url="${ruby_url:-$ruby_mirror/vm-$ruby_version/$ruby_archive}"
 #
 function install_ruby()
 {
+	if [[ "$install_dir" == '/usr/local' ]]; then
+		error "Unsupported see https://github.com/oracle/truffleruby/issues/1389"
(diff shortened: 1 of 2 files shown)


eregon (approved):
Looks good to me.

I guess technically one could use other system-wide locations (e.g., /usr), but probably nobody does in practice?

postmodern:
Thoughts on moving the `fail` into the `install_ruby` function, at the top?

missingcharacter:
@postmodern like this (see new commit) or turn it into an `if`?

postmodern:
@missingcharacter yeah `if` would be better since we're only doing one comparison.

Also, since we're in the `install_ruby` function, we could also use `error` and `return 1` to abort back to [bin/ruby-install](https://github.com/postmodern/ruby-install/blob/257746199f83794294326383d4a3ac8f37ab9d4a/bin/ruby-install#L50) where a generic error message would also be printed. You're call on whether to print one error message and immediately exit, or print two error messages.

missingcharacter:
@postmodern how about now?

postmodern:
Squash merged, even though GitHub Actions keeps failing shunit2 test suites only on macOS; still need to investigate that.

PR #402, opened 2021-06-04
Update the TruffleRuby + GraalVM CE distribution to use Java 11.

nirvdrum:
GraalVM Community Edition (CE), which bundles its own version of Java, ships with various Java versions as its base. Java 11 is the latest long-term support and the recommended version to use. Incidentally, there are no Java 8-based GraalVM CE builds for macOS, so moving to the Java 11 base allows installation on macOS as well.


eregon (approved):
Thanks!

Issue #409, opened 2021-09-24
Can't install ruby on M1 Mac  "use of undeclared identifier 'RSA_SSLV23_PADDING'"


SteveTro:
## Steps To Reproduce

Steps to reproduce the bug:
1. `$ ruby-install ruby -- --with-openssl-dir=$(brew --prefix openssl)`

## Expected Behavior

ruby should be installed

## Actual Behavior

```
ossl_pkey_rsa.c:950:5: error: use of undeclared identifier 'RSA_SSLV23_PADDING'
    DefRSAConst(SSLV23_PADDING);
    ^
ossl_pkey_rsa.c:885:58: note: expanded from macro 'DefRSAConst'
#define DefRSAConst(x) rb_define_const(cRSA, #x, INT2NUM(RSA_##x))
                                                         ^
<scratch space>:166:1: note: expanded from here
RSA_SSLV23_PADDING
^
110 warnings and 1 error generated.
make[2]: *** [ossl_pkey_rsa.o] Error 1
make[1]: *** [ext/openssl/all] Error 2
make: *** [build-ext] Error 2
```

## Environment

    $ ruby-install --version
    ruby-install: 0.8.2
    $ uname -a
    Darwin macbook.fritz.box 20.3.0 Darwin Kernel Version 20.3.0: Thu Jan 21 00:06:51 PST 2021; root:xnu-7195.81.3~1/RELEASE_ARM64_T8101 arm64
    $ cc --version
    Apple clang version 12.0.5 (clang-1205.0.22.9)
    Target: arm64-apple-darwin20.3.0
    Thread model: posix
    InstalledDir: /Library/Developer/CommandLineTools/usr/bin
    $ openssl version
    OpenSSL 3.0.0 7 sep 2021 (Library: OpenSSL 3.0.0 7 sep 2021)

superx3man:
Same issue observed for intel processor as well.

postmodern:
This looks like a compatibility difference between the ruby and the version of openssl. Also, `- --with-openssl-dir=$(brew --prefix openssl)` is not necessary, as ruby-install already passes in `--with-opt-dir="$(brew --prefix openssl):$(brew --prefix readline):$(brew --prefix libyaml):$(brew --prefix gdbm)"`.

Can you try reproducing this without ruby-install?

```
$ wget https://cache.ruby-lang.org/pub/ruby/3.0/ruby-3.0.2.tar.xz
[...]

PR #410, opened 2021-09-24
Force homebrew to use OpenSSL 1.1

emptyflask:
OpenSSL 3.0 is released, and Homebrew uses it by default now, but Ruby doesn't support it yet.
See the issue at https://github.com/postmodern/ruby-install/issues/409


emptyflask:
Tested with the latest ruby and truffleruby. rbx seems broken for an unrelated reason, but I'd be surprised if it's expecting a newer openssl.

postmodern:
Looks good. Unpinning openssl when rubies start supporting OpenSSL 3.0 may be tricky. OpenSSL 1.1 appears to still be maintained for now, so we don't have to worry about missing any security updates if we continue to target 1.1. Eventually we will want to unpin and switch back to 3.0.

postmodern:
ruby-install 0.8.3 has been released. Now we wait for homebrew to pick it up.

aduh95:
https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/ lists that they have updated OpenSSL to 3.0.0, won't that conflict with this change?

dgholz:
@

> https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/ lists that they have updated OpenSSL to 3.0.0, won't that conflict with this change?

That's the `openssl` gem's version, which tends to follow the max OpenSSL version it supports. It still works fine with OpenSSL `1.1`

PR #425, opened 2022-05-26
Update README to reflect new ruby-install port

htp:
Hello!

I noticed the FreeBSD port linked in the README seemed to be stuck at version 0.6.0 and [had issues building on newer FreeBSDs](https://github.com/steakknife/ruby-install-freebsd/issues/6), so I took the liberty of putting together an updated port and submitting it for inclusion in the official port collection.

I'm happy to share [the port was accepted](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=264168), so here's a corresponding change to ruby-install's README to point folks to the now-official port. Thanks for chruby and ruby-install!

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,8 +165,9 @@ ruby-install is available on [Fedora Copr](https://copr.fedorainfracloud.org/cop
 
 ### FreeBSD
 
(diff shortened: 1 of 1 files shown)


postmodern:
Awesome! Great work getting `ruby-install` into the main repo.

PR #428, opened 2022-07-29
Add support for AArch64 builds of TruffleRuby.

nirvdrum:
TruffleRuby 22.2.0 is the first release to support the new ARM-based Apple machines. AArch64 on Linux has also been present for a while. This PR allows TruffleRuby and the TruffleRuby+GraalVM builds to install on non-AMD64 platforms.


eregon (commented):
Looks good to me.

PR #443, opened 2023-02-21
Fix Shellcheck violations

### LINKED ISSUE #442: make check is failing
## Description

Running `make check` fails

## Steps To Reproduce

Steps to reproduce the bug:
1. clone the repo
2. run `make check`

## Expected Behavior

All checks should pass

## Actual Behavior

Fails with lots of shellcheck errors

ixti:
Resolves: #442

### DIFF
--- a/share/ruby-install/functions.sh
+++ b/share/ruby-install/functions.sh
@@ -19,7 +19,10 @@ function load_dependencies_from()
 {
 	local file="$1"
 
-	ruby_dependencies=($(fetch "$ruby/$file" "$package_manager" || return $?))
+	ruby_dependencies=()
+	while IFS='' read -r line; do
+		ruby_dependencies+=("$line")
+	done < <(fetch "$ruby/$file" "$package_manager" || return $?)
 }
 
 #
--- a/share/ruby-install/package_manager.sh
+++ b/share/ruby-install/package_manager.sh
@@ -29,7 +29,8 @@ function install_packages()
 			run $sudo pkg install -y "$@" || return $?
 			;;
 		brew)
-			local brew_owner="$(/usr/bin/stat -f %Su "$(command -v brew)")"
+			local brew_owner=
+			brew_owner="$(/usr/bin/stat -f %Su "$(command -v brew)")"
 			local brew_sudo=""
 
 			if [[ "$brew_owner" != "$(id -un)" ]]; then
--- a/share/ruby-install/package_manager.sh
+++ b/share/ruby-install/package_manager.sh
@@ -41,7 +41,10 @@ function install_packages()
 			run $brew_sudo brew upgrade "$@" || return $?
 			;;
 		pacman)
-			local missing_pkgs=($(pacman -T "$@"))
+			local missing_pkgs=()
+			while IFS='' read -r line; do
+				missing_pkgs+=("$line")
+			done < <(pacman -T "$@")
 
 			if (( ${#missing_pkgs[@]} > 0 )); then
 				run $sudo pacman -S "${missing_pkgs[@]}" || return $?
--- a/share/ruby-install/util.sh
+++ b/share/ruby-install/util.sh
@@ -12,7 +12,7 @@ function fetch()
 
 	while IFS="" read -r line; do
 		if [[ "$line" == "$key:"* ]]; then
-			echo "${line##$key:*([[:space:]])}"
+			echo "${line##"$key":*([[:space:]])}"
 		fi
 	done < "$file"
 }
--- a/share/ruby-install/checksums.sh
+++ b/share/ruby-install/checksums.sh
@@ -30,7 +30,8 @@ function lookup_checksum()
 		return 1
 	fi
 
-	local output="$(grep "  $file" "$checksums")"
+	local output
+	output="$(grep "  $file" "$checksums")"
 
 	echo -n "${output%% *}"
 }
@@ -55,7 +56,8 @@ function compute_checksum()
 	fi
 
 	debug "$program $file"
-	local output="$($program "$file")"
+	local output
+	output="$($program "$file")"
 
 	echo -n "${output%% *}"
 }
@@ -71,7 +73,8 @@ function verify_checksum()
 		return
 	fi
 
-	local actual_checksum="$(compute_checksum "$algorithm" "$file")"
+	local actual_checksum
+	actual_checksum="$(compute_checksum "$algorithm" "$file")"
 
 	if [[ "$actual_checksum" != "$expected_checksum" ]]; then
 		error "Invalid $algorithm checksum for $file"
--- a/share/ruby-install/ruby-install.sh
+++ b/share/ruby-install/ruby-install.sh
@@ -272,7 +272,8 @@ function init()
 		return 1
 	fi
 
-	local fully_qualified_version="$(lookup_ruby_version "$ruby" "$ruby_version")"
+	local fully_qualified_version
+	fully_qualified_version="$(lookup_ruby_version "$ruby" "$ruby_version")"
 
 	if [[ -n "$fully_qualified_version" ]]; then
 		ruby_version="$fully_qualified_version"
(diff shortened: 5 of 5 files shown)


share/ruby-install/util.sh
```diff
@@ -12,7 +12,7 @@ function fetch()
 
 	while IFS="" read -r line; do
 		if [[ "$line" == "$key:"* ]]; then
-			echo "${line##$key:*([[:space:]])}"
+			echo "${line##"$key":*([[:space:]])}"
```
postmodern, line 15: Are quotes really needed within bash variable substitutions?

share/ruby-install/functions.sh
```diff
@@ -19,7 +19,10 @@ function load_dependencies_from()
 {
 	local file="$1"
 
-	ruby_dependencies=($(fetch "$ruby/$file" "$package_manager" || return $?))
+	ruby_dependencies=()
+	while IFS='' read -r line; do
+		ruby_dependencies+=("$line")
+	done < <(fetch "$ruby/$file" "$package_manager" || return $?)
```
postmodern, line 25: We actually need the implicit shell-splitting here. `fetch "$ruby/$file" "$package_manager"` will read the `dependencies.txt` file and return the `apt: pkg1 pkg2 ...` line that matches the `"$package_manager"`, but without the `apt: ` prefix. It is then implicitly splatted into an Array of individual package names. Reading each line would cause `ruby_dependencies` to become a singleton Array that contains the whole line (ex: `("pkg1 pkg2 pkg3")`).

share/ruby-install/package_manager.sh
```diff
@@ -41,7 +41,10 @@ function install_packages()
 			run $brew_sudo brew upgrade "$@" || return $?
 			;;
 		pacman)
-			local missing_pkgs=($(pacman -T "$@"))
+			local missing_pkgs=()
+			while IFS='' read -r line; do
+				missing_pkgs+=("$line")
```
postmodern, line 46: I'm not a Arch user so I can't verify this, can you double check if `pacman -T` outputs the missing packages one per-line or a single-line with the package names separated by spaces?
https://archlinux.org/pacman/pacman.8.html

postmodern (changes requested):
Spotted a few things. We may want to disable a few more of the ShellCheck rules, since they might not correctly be matching the code.

postmodern:
Looking at some of the ShellCheck errors, I think `SC2155` might be wrong here:

```
In share/ruby-install/checksums.sh line 33:
	local output="$(grep "  $file" "$checksums")"
              ^----^ SC2155: Declare and assign separately to avoid masking return values.


In share/ruby-install/checksums.sh line 58:
	local output="$($program "$file")"
              ^----^ SC2155: Declare and assign separately to avoid masking return values.


In share/ruby-install/checksums.sh line 74:
	local actual_checksum="$(compute_checksum "$algorithm" "$file")"
              ^-------------^ SC2155: Declare and assign separately to avoid masking return values.


In share/ruby-install/functions.sh line 22:
	ruby_dependencies=($(fetch "$ruby/$file" "$package_manager" || return $?))
                           ^-- SC2207: Prefer mapfile or read -a to split command output (or quote to avoid splitting).


In share/ruby-install/package_manager.sh line 32:
			local brew_owner="$(/usr/bin/stat -f %Su "$(command -v brew)")"
                              ^--------^ SC2155: Declare and assign separately to avoid masking return values.


In share/ruby-install/package_manager.sh line 43:
			local missing_pkgs=($(pacman -T "$@"))
                                            ^---------------^ SC2207: Prefer mapfile or read -a to split command output (or quote to avoid splitting).


In share/ruby-install/ruby-install.sh line 275:
	local fully_qualified_version="$(lookup_ruby_version "$ruby" "$ruby_version")"
              ^---------------------^ SC2155: Declare and assign separately to avoid masking return values.
```

When you put a command inside of `$()` the shell will implicitly ignore any returned error codes since it only cares about the output of the command. We might be able to just exclude `SC2155` as well.

share/ruby-install/util.sh
ixti: I don't think so, but shellcheck disagrees :D

share/ruby-install/functions.sh
ixti: Hmm, agree. I just followed (blindly) what Shellcheck WIKI was proposing :D

share/ruby-install/package_manager.sh
ixti: Don't have Arch installed, either. I'm just maintaining Gentoo ebuild in [::guru overlay](https://github.com/gentoo/guru/tree/dev/dev-ruby/ruby-install), and currently I simply disable `make check` :D

ixti:
I'm fine if those checks will be disabled (globally or locally via comments). Just went the lazy approach by doing what SC proposes and making sure tests pass after that on my machine :D

share/ruby-install/util.sh
postmodern: Wait, which version of ShellCheck are you using? I'm on 0.7.2, which might be older than your version, which might explain why I'm not seeing those warnings?

share/ruby-install/package_manager.sh
postmodern: Wait, I forgot I can check using the [archlinux docker image](https://hub.docker.com/_/archlinux).

share/ruby-install/util.sh
ixti: Mine is 0.8.0 :D

ixti:
This PR got massively outdated, LOL.

Current HEAD still has issues with Shellcheck 0.11, but from what I can tell - all of them most likely should be ignored:

<details><summary>make check</summary>

```
❯ make check
shellcheck --exclude SC2034 \
           --exclude SC2154 \
           --exclude SC1090 \
           --exclude SC1091 \
           --exclude SC2242 \
           share/ruby-install/*.sh bin/*

In share/ruby-install/checksums.sh line 33:
        local output="$(grep "  $file" "$checksums")"
              ^----^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In share/ruby-install/checksums.sh line 58:
        local output="$($program "$file")"
              ^----^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In share/ruby-install/checksums.sh line 74:
        local actual_checksum="$(compute_checksum "$algorithm" "$file")"
              ^-------------^ SC2155 (warning): Declare and assign separately to avoid masking return values.


In share/ruby-install/functions.sh line 69:
                        patches[$i]="$dest"
                                ^-- SC2004 (style): $/${} is unnecessary on arithmetic variables.


In share/ruby-install/package_manager.sh line 37:
                                brew_sudo="sudo -u \"$SUDO_USER\""
                                           ^--------^ SC2089 (warning): Quotes/backslashes will be treated literally. Use an array.


In share/ruby-install/package_manager.sh line 40:
                        run $brew_sudo brew install "$@" ||
                            ^--------^ SC2090 (warning): Quotes/backslashes in this variable will not be respected.


In share/ruby-install/package_manager.sh line 41:
                        run $brew_sudo brew upgrade "$@" || return $?
                            ^--------^ SC2090 (warning): Quotes/backslashes in this variable will not be respected.


In share/ruby-install/package_manager.sh line 44:
                        local missing_pkgs=($(pacman -T "$@"))
                                            ^---------------^ SC2207 (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).


In share/ruby-install/ruby-install.sh line 90:
                        ruby_version="${string#$ruby-}"
                                               ^---^ SC2295 (info): Expansions inside ${..} need to be quoted separately, otherwise they match as patterns.

Did you mean:
                        ruby_version="${string#"$ruby"-}"


In share/ruby-install/ruby-install.sh line 284:
        local fully_qualified_version="$(lookup_ruby_version "$ruby" "$ruby_version")"
              ^---------------------^ SC2155 (warning): Declare and assign separately to avoid masking return values.

For more information:
  https://www.shellcheck.net/wiki/SC2089 -- Quotes/backslashes will be treate...
  https://www.shellcheck.net/wiki/SC2090 -- Quotes/backslashes in this variab...
  https://www.shellcheck.net/wiki/SC2155 -- Declare and assign separately to ...
make: *** [Makefile:49: check] Error 1
```
</details>

PR #453, opened 2023-06-19
Add TruffleRuby and TruffleRuby+GraalVM 23.0.0

eregon:
* Use the new Oracle GraalVM distribution which uses the GFTC license as it is significantly faster than GraalVM CE and is free for development and production use.
* See https://medium.com/graalvm/whats-new-in-graalvm-languages-161527df3d76

ruby-versions PR: https://github.com/postmodern/ruby-versions/pull/70

### DIFF
--- a/share/ruby-install/truffleruby-graalvm/functions.sh
+++ b/share/ruby-install/truffleruby-graalvm/functions.sh
@@ -13,10 +13,19 @@ case "$os_arch" in
 	*)       fail "Unsupported platform $os_arch" ;;
 esac
 
-ruby_dir_name="graalvm-ce-java11-$ruby_version"
-ruby_archive="${ruby_archive:-graalvm-ce-java11-$graalvm_platform-$graalvm_arch-$ruby_version.tar.gz}"
-ruby_mirror="${ruby_mirror:-https://github.com/graalvm/graalvm-ce-builds/releases/download}"
-ruby_url="${ruby_url:-$ruby_mirror/vm-$ruby_version/$ruby_archive}"
+if [ "$ruby_version" = "23.0.0" ]; then
+	log "TruffleRuby-GraalVM 23.0 and later installed by ruby-install use the faster Oracle GraalVM distribution"
+	log "Oracle GraalVM uses the GFTC license, which is free for development and production use, see https://medium.com/graalvm/161527df3d76"
+	ruby_dir_name="graalvm-jdk-17.0.7+8.1"
+	ruby_archive="${ruby_archive:-graalvm-jdk-17.0.7_${graalvm_platform/darwin/macos}-${graalvm_arch/amd64/x64}_bin.tar.gz}"
+	ruby_mirror="${ruby_mirror:-https://download.oracle.com/graalvm/17/archive}"
+	ruby_url="${ruby_url:-$ruby_mirror/$ruby_archive}"
+else
+	ruby_dir_name="graalvm-ce-java11-$ruby_version"
+	ruby_archive="${ruby_archive:-graalvm-ce-java11-$graalvm_platform-$graalvm_arch-$ruby_version.tar.gz}"
+	ruby_mirror="${ruby_mirror:-https://github.com/graalvm/graalvm-ce-builds/releases/download}"
+	ruby_url="${ruby_url:-$ruby_mirror/vm-$ruby_version/$ruby_archive}"
+fi
 
 #
 # Install GraalVM into $install_dir.
--- a/share/ruby-install/truffleruby/functions.sh
+++ b/share/ruby-install/truffleruby/functions.sh
@@ -15,8 +15,24 @@ esac
 
 ruby_dir_name="truffleruby-$ruby_version-$truffleruby_platform-$truffleruby_arch"
 ruby_archive="${ruby_archive:-$ruby_dir_name.tar.gz}"
-ruby_mirror="${ruby_mirror:-https://github.com/oracle/truffleruby/releases/download}"
-ruby_url="${ruby_url:-$ruby_mirror/vm-$ruby_version/$ruby_archive}"
+
+if [ "$ruby_version" = "23.0.0" ]; then
+	log "TruffleRuby 23.0 and later installed by ruby-install use the faster Oracle GraalVM distribution"
+	log "Oracle GraalVM uses the GFTC license, which is free for development and production use, see https://medium.com/graalvm/161527df3d76"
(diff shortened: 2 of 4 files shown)


share/ruby-install/truffleruby-graalvm/functions.sh
```diff
@@ -13,10 +13,19 @@ case "$os_arch" in
...
 
-ruby_dir_name="graalvm-ce-java11-$ruby_version"
-ruby_archive="${ruby_archive:-graalvm-ce-java11-$graalvm_platform-$graalvm_arch-$ruby_version.tar.gz}"
-ruby_mirror="${ruby_mirror:-https://github.com/graalvm/graalvm-ce-builds/releases/download}"
-ruby_url="${ruby_url:-$ruby_mirror/vm-$ruby_version/$ruby_archive}"
+if [ "$ruby_version" = "23.0.0" ]; then
+	log "TruffleRuby-GraalVM 23.0 and later installed by ruby-install use the faster Oracle GraalVM distribution"
+	log "Oracle GraalVM uses the GFTC license, which is free for development and production use, see https://medium.com/graalvm/161527df3d76"
+	ruby_dir_name="graalvm-jdk-17.0.7+8.1"
+	ruby_archive="${ruby_archive:-graalvm-jdk-17.0.7_${graalvm_platform/darwin/macos}-${graalvm_arch/amd64/x64}_bin.tar.gz}"
```
eregon, line 20: Unfortunately I don't know of any way to deduce `jdk-17.0.7+8.1` and `jdk-17.0.7` from the GraalVM/TruffleRuby version 23.0.0.
This different versioning is due to aligning with JDK releases, including the [Archive URLs](https://www.oracle.com/java/technologies/jdk-script-friendly-urls/).

At some point the versions might become the same (around JDK 25) or be more correlated, but for now they are separate. This is not entirely figured out yet, TruffleRuby might also get its own separate version.

I think ideally `ruby-versions`
* would also contain URLs (per ruby, version, os, arch), a bit like ruby-build.
* `ruby_dir_name` wouldn't need to be the dir name inside the archive but be automatically deduced/not need to be known, or part of ruby-versions

Then we wouldn't need to update ruby-install. Until ruby-versions supports that we'll have to update ruby-install.

share/ruby-install/truffleruby/functions.sh
```diff
@@ -15,8 +15,24 @@ esac
...
-ruby_mirror="${ruby_mirror:-https://github.com/oracle/truffleruby/releases/download}"
-ruby_url="${ruby_url:-$ruby_mirror/vm-$ruby_version/$ruby_archive}"
+
+if [ "$ruby_version" = "23.0.0" ]; then
+	log "TruffleRuby 23.0 and later installed by ruby-install use the faster Oracle GraalVM distribution"
+	log "Oracle GraalVM uses the GFTC license, which is free for development and production use, see https://medium.com/graalvm/161527df3d76"
+	ruby_mirror="${ruby_mirror:-https://gds.oracle.com/api/20220101/artifacts}"
+	truffleruby_artifact_id=""
+	case "$truffleruby_platform-$truffleruby_arch" in
+		linux-amd64)    truffleruby_artifact_id="FD4AB182EA4CEDFDE0531518000AF13E" ;;
```
eregon, line 25: These URLs are not pretty, but they are stable and should keep working forever.
We're working on getting a stable URL format for them, there are lots of changes here with the new distribution and aligning with JDK releases.

share/ruby-install/truffleruby-graalvm/functions.sh
postmodern: I'd suggest just picking the most optimal version based on what JDK versions are currently available via the various package managers (homebrew, `apt`, `dnf`).

eregon: > I'd suggest just picking the most optimal version based on what JDK versions are currently available via the various package managers (homebrew, `apt`, `dnf`).

What is downloaded here is a GraalVM JDK, not just "a JDK" or OpenJDK (for example it contains `bin/gu`, the GraalVM Updater, used to install the truffleruby component).
So it's not available on all package managers (at least not `apt`).

There is a URL for "latest 17" which is https://download.oracle.com/graalvm/17/latest/graalvm-jdk-17_linux-x64_bin.tar.gz but then the SHA sum would change when there is a new patch release of 17, which seems not what we want (notably it wouldn't match the SHA in ruby-versions anymore).

Instead, TruffleRuby GraalVM 23.0.1 would use jdk 17.0.8 and be a separate release.

> ruby-versions will not and should not contain dependency information as that is scope creep.

OK, then it means we need to change ruby-install for every truffleruby release until we can deduce these 2 version numbers one from another. I am OK with it, I was just trying to think of ways to resolve that.

eregon: BTW, there are plans to start with the TruffleRuby standalone and then later install the GraalVM JDK into it to enable `--jvm` mode. Once that work is done then we shouldn't need to know the URL of the corresponding GraalVM JDK in Ruby installers anymore.

eregon:
This and https://github.com/postmodern/ruby-versions/pull/70 should be ready now, I added libyaml to the dependencies, that's new since 23.0.

postmodern (changes requested):
Now that [your PR to `ruby-versions`](https://github.com/postmodern/ruby-versions/pull/70) has been merged `ruby_versions_url` can be switched back.

share/ruby-install/ruby-versions.sh
```diff
@@ -3,7 +3,7 @@
 source "$ruby_install_dir/versions.sh"
 source "$ruby_install_dir/checksums.sh"
 
-ruby_versions_url="https://raw.githubusercontent.com/postmodern/ruby-versions/master"
+ruby_versions_url="https://raw.githubusercontent.com/eregon/ruby-versions/truffleruby-23.0.0"
```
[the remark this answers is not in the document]
eregon: Done

share/ruby-install/truffleruby-graalvm/functions.sh
eregon: > We could host the file in the `ruby-install` repository and download it using the `download` built-in function, just like how we download the `ruby-versions` files. Could potentially also do the same with the SHA checksums for the GraalVM JDK.

Right. I'd like to keep it simple for now because it's not clear yet how it will look like for future releases, probably it will be significantly simpler.

postmodern (approved):

PR #454, opened 2023-06-21
Fix macos handling for truffleruby artifacts

rwstauner:
We change darwin to macos at the top of this file.

I confirmed that on my mac i can install
- truffleruby-23.0.0
- truffleruby-22.3.0
- truffleruby-graalvm-23.0.0
- truffleruby-graalvm-22.3.0


postmodern (approved):

postmodern:
@eregon any other testing or fixes, or should I start preparing a 0.9.1 release?

nirvdrum:
@postmodern I think we're all set now. It was an unfortunate copy & paste issue. @rwstauner has verified the macOS builds install correctly and @eregon had confirmed the Linux ones do as well. Whenever GitHub adds support for macOS ARM machines we may do well to add automated tests that check the tarball downloads work.

eregon:
Yes, all good, a release would be great :)

postmodern:
Will try to do that tomorrow, after the dentist appointment...

PR #459, opened 2023-07-14
Use apt-get instead of apt

joshheinrichs-shopify:
When running non-interactively, this message appears in the logs:

    >>> Installing dependencies for ruby 3.2.1 ...
    WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

This isn't a huge deal, but apt doesn't appear to respect some options in /etc/apt/apt.conf.d/ when ran non-interactively. These options are respected when running apt-get. This is at least true for DPkg::Lock::Timeout, but may apply to other options as well.

### DIFF
--- a/share/ruby-install/package_manager.sh
+++ b/share/ruby-install/package_manager.sh
@@ -17,7 +17,7 @@ function install_packages()
 {
 	case "$package_manager" in
 		apt)
-			run $sudo apt install -y "$@" || return $?
+			run $sudo apt-get install -y "$@" || return $?
 			;;
 		dnf|yum)
 			run $sudo "$package_manager" install -y "$@" || return $?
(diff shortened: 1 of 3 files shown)


joshheinrichs-shopify:
I dug into this a bit deeper and it looks like `apt` and `apt-config` set their lock timeout based on `Binary::apt::Dpkg::Lock::Timeout` rather than `DPkg::Lock::Timeout`.

https://salsa.debian.org/apt-team/apt/-/blob/a19f606aad717fe5c9c69237c3af53feb547115e/apt-private/private-cmndline.cc#L498-501

It's still probably better to use apt-get than apt, but we don't absolutely need this if apt's output is highly desirable in interactive contexts.

postmodern:
One of the benefits of `apt` is it has a better ANSI progress bar and friendlier output. Maybe we could fallback to `apt-get` if stdout is not a TTY (ex: `if [[ -t 1 ]]; then`).

joshheinrichs-shopify:
👍 if this becomes a common theme, switching based on whether we're a tty / a env var or something makes sense, but for the sake of simplicity i'm happy to stick with a single front end since there's a workaround for my specific problem.

Issue #464, opened 2023-09-05
doesn't make some simple yet very important checks before starting the installation process (e.g need permission when using --sytem flag)


Mizou04:
## Steps To Reproduce

Steps to reproduce the bug:
1. `$ ruby-install ...`
2. ???

## Environment

    $ ruby-install --version
    ...
    $ uname -a
    ...
    $ cc --version
    ...
    $ openssl version
    ...

Mizou04:
I have a fix for this issue

postmodern:
@Mizou04 feel free to submit a PR. I'm guessing an extra `if [[ -w "$install_dir" ]]; then` could be added to `bin/ruby-install`.

Mizou04:
actually I modified directly the function responsible for parsing the options so it doesn't need to go the whole stack down to bin/ruby-install, is that ok? also I couldn't open a PR

postmodern:
Going to close since I couldn't come up with a solution that covers all edge cases (ex: `ruby-install --install-dir /usr/local`, `ruby-install --install-dir /some/other/dir`, or when `/usr/local` is writable by a special user group that the user belongs to or a file-system ACL, etc). Even if I wrote a function which checks if each parent directory is writable, that could still be fooled by symlinks. Plus, [ruby-build](https://github.com/rbenv/ruby-build) does check permissions before compiling either. I ended up just changing the example in the `README` to be `sudo ruby-install --system ...`.

postmodern:
Ended up adding a basic permissions check that recursively tests the installation directory's parent directories for writability (see eeb4575). This *will* catch when you run `ruby-install --system` without `sudo`. This will be released in 0.10.0.

PR #465, opened 2023-09-05
warn user that installation with --system flag will fail if she's not root #464

Mizou04:
fix for issue #464

### DIFF
--- a/share/ruby-install/ruby-install.sh
+++ b/share/ruby-install/ruby-install.sh
@@ -106,6 +106,20 @@ function parse_ruby()
 }
 
 #
+#Called when selecting --system flag                                                     
+#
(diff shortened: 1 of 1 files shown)


postmodern (changes requested):
It's generally a bad practice to put additional logic in option parsing. The option parser should only parse the options. Any validation logic should happen once *all* of the options have been parsed. This prevents issues where options can be specified in different orders.

I feel like testing if `$install_dir` is writable is a better solution, as that will also catch edge-cases where the user is trying to explicitly install ruby to somewhere else on the system: `ruby-install --install-dir /opt/ ruby`, etc This guard logic could be added to `pre_install` or directly to `bin/ruby-install`.

PR #466, opened 2023-09-05
warn user that installation will fail if she doesn't have necessary write permissions

Mizou04:
added the guard logic against absent write permissions in `bin/ruby-install`

### DIFF
--- a/test/cli-tests/warn_if_no_write_permissions_test.sh
+++ b/test/cli-tests/warn_if_no_write_permissions_test.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+. ./test/helper.sh
+
+test_install_dir="$test_fixtures_dir/no_write_permissions_test"
+
+function setUp()
+{
+	mkdir -p "$test_install_dir/bin"
+	chmod -w "$test_install_dir/bin"
+}
+
+function test_no_write_permissions()
+{
+	local output status
+	output="$(ruby-install --install-dir "$test_install_dir/bin/rubies" ruby 2>&1)"
+	status=$?
+
+	assertEquals "did not return 0" 255 $status
+	assertTrue "did not print a message to STDOUT" \
+		'[[ "$output" == *"Installation directory is not writable by the user"* ]]'
+}
+
+function tearDown()
+{
+	rm -rf "$test_install_dir"
+}
+
+SHUNIT_PARENT=$0 . $SHUNIT2
--- a/share/ruby-install/functions.sh
+++ b/share/ruby-install/functions.sh
@@ -7,8 +7,22 @@ source "$ruby_install_dir/checksums.sh"
 #
 function pre_install()
 {
+
+	local status
+	local exisiting_part_dir="${install_dir%/*}"
+
 	mkdir -p "$src_dir" || return $?
-	mkdir -p "${install_dir%/*}" || return $?
+
+	while [ ! -d "$exisiting_part_dir" ]; do
+		exisiting_part_dir="${exisiting_part_dir%/*}"
+	done
+
+	if [ ! -w "$exisiting_part_dir" ]; then
+		fail "Installation directory is not writable by the user : ${exisiting_part_dir}"
+		status=$?
+	fi
+
+	mkdir -p "${install_dir%/*}" || return $status
 }
 
 #


bin/ruby-install
```diff
@@ -18,6 +18,11 @@ fi
 
 init || exit $?
 
+if [[ -w "$install_dir" ]]; then
```
postmodern, line 21: Should be `! -w` for not writable.

bin/ruby-install
```diff
@@ -18,6 +18,11 @@ fi
...
+	error "seems you don't have necessary write permission on \"$install_dir\""
```
postmodern, line 22: Quotes in the error message are unnecessary.

postmodern (changes requested):
Minor changes requested. `git commit --amend` and force push the branch, instead of closing the PR.

Also, this needs unit-tests. Luckily, you can copy [`test/cli-tests/no_reinstall_test.sh`](https://github.com/postmodern/ruby-install/blob/master/test/cli-tests/no_reinstall_test.sh) and adapt the test file. The test file just needs to do `chmod -w $test_install_dir` in the `setUp()` function before the tests.

bin/ruby-install
Mizou04: 🤦‍♂️😅

test/cli-tests/no_permission_warn_test.sh
```diff
@@ -0,0 +1,27 @@
...
+{
+	mkdir -p $test_install_dir
+	chmod -w $test_install_dir
+}
+
+function test_no_install_when_write_permissions_lacking()
+{
+	local output="$(ruby-install --install-dir "$test_install_dir" ruby)"
+
+	assertEquals "did not return 0" 0 $?
```
postmodern, line 17: Should test that `$?` is `1` since we're exiting with `1`.

bin/ruby-install
```diff
@@ -18,6 +18,11 @@ fi
 
 init || exit $?
 
+if [[ ! -w "$install_dir" ]]; then
+	log "seems like you need necessary write permission on $install_dir"
```
postmodern, line 22: Error message should be more authoritative. `"installation directory is not writable by the user: $install_dir"`

test/cli-tests/no_permission_warn_test.sh
```diff
@@ -0,0 +1,27 @@
...
+	assertTrue "did not print a message to STDOUT" \
+		'[[ "$output" == *"need necessary write permission"* ]]'
```
postmodern, line 19: Match the full error message.

postmodern (changes requested):
Minor changes.

bin/ruby-install
Mizou04: I see

postmodern: Also use `error "..."` or `fail "..."` (which also does a `exit -1`) instead of `log`, as this is an error message.

test/cli-tests/no_permission_warn_test.sh
Mizou04: I think there's an error with the test when exiting with 1 status

postmodern: Try doing:

```shell
local ret
local output="$(...)" || ret=$?

assertEquals "..." -1 $ret
```

That should safely capture the exit status.

Mizou04: this would return the exit status of setting the variable `local output`, if `ret` is not declared before, and if it was declared with `local ret` it will print hold an `empty value`, the solution is a little tricky, example:
```bash
function err(){                                                                           
  echo "bye"                                                                              
  exit 3                                                                                  
}                                                                                       
                                                                                             
function setting(){                                                                       
  local output                                                                            
  local e                                                                                 
  #Z5Z5Z is for uniqness                                                                  
  inter="$(err)Z5Z5Z$?"                                                                   
  output="$(echo $inter | awk 'BEGIN { FS="Z5Z5Z" } { print $1 }')"                       
       e="$(echo $inter | awk 'BEGIN { FS="Z5Z5Z" } { print $2 }')"                       
    echo $output  #will print "bye"                                                                            
    echo $e       #will print 3 (the exit status of err() function)                                                                                 
 }
```

Mizou04: also the fail logging function exit with status -1, but in Unix the status codes are unsigned 8-bit integers between 0 and 255, so `exit -1` becomes `exit 255`, should I open new issue for it?

postmodern: Ah, you are correct. Yes compare against `255`.

test/cli-tests/no_permission_warn_test.sh
```diff
@@ -0,0 +1,32 @@
...
+{
+	mkdir -p "$test_install_dir"
+	chmod -w "$test_install_dir"
+}
+
+function test_no_install_when_write_permissions_lacking()
+{
+	local output
+	local status
+	local inter="$(ruby-install --install-dir "$test_install_dir" ruby 2>&1)Z5Z5Z$?"
```
postmodern, line 17: I think this is too complex. `local output="$(...)"; ret=$?` should capture both successful and failure exit statuses.

postmodern (changes requested):

test/cli-tests/no_permission_warn_test.sh
Mizou04: I've gone with  `error()` instead (checkout last commit), I think it's better to change the fail() exit status so it become obvious what status is exiting with.

test/cli-tests/no_permission_warn_test.sh
Mizou04: with this method, `ret` catches the exit status of the assignment rather than the command itself. the redirecting to stdout is also necessary for catching the error message

postmodern:
```shell
output="$(exit -1)"; ret=$?
echo $ret
# 255
output="$(exit 0)"; ret=$?
echo $ret
# 0
```

Both `output="$(...)" || ret=$?` and `output="$(...)"; ret=$?` should capture the exit status, because the exit status of the previous command is kept in `$?`, until the current command or any embedded sub-command is completely executed.

Mizou04: yes, if we dropped `local` modifier and declared output as global

postmodern: Oh right, `local` has it's own implicit return value. In that case the solution is simple, split the `local` definition from the assignment:

```shell
test_exit_status() {
	local output
	local ret

	output="$(exit -1)"
	ret=$?

	if [[ $ret == 255 ]]; then
		echo "exit status = -1"
	fi
}

test_exit_status
# exit status = -1
```

Mizou04:
Can you review the last commit?

bin/ruby-install
```diff
@@ -18,6 +18,10 @@ fi
 
 init || exit $?
 
+if [[ ! -w "$install_dir" ]]; then
```
postmodern, line 21: I just realized an edge-case. If `$install_dir` doesn't exist yet, such as when running `ruby-install` for the first time on a new system that doesn't have `/opt/rubies` or `~/.rubies` yet, it will consider the `$install_dir` non-writable. We should also check for it's existence *and* if it's not-writable.

```shell
if [[ -d "$install_dir" ]] && [[ ! -w "$install_dir" ]]; then
  ...
fi
```

postmodern (changes requested):
Just thought of an edge-case where this would break for new users that do not have `/opt/rubies` or `~/.rubies` on their system's yet.

bin/ruby-install
postmodern: This might cause problems if the user specifies a really long `--install-dir` who's parent directories do not exist yet.

```
ruby-instlal --install-dir `$HOME/does/not/exist/yet` ruby
```

We could move this logic into `share/ruby-install/functions.sh` `pre_install` which already does a `mkdir -p "${install_dir%/*}" || return $?`.

Mizou04: oh yes, that was close!!!

Mizou04: there's an issue I think with this, for example if `install_dir` is of this form : `/this/part/exists/this/part/exists/not`, we should recursively (iteratively) return backwards to check for the existing part, before testing if it was writable or not; then we proceed with our normal operation

postmodern: Or we could take advantage of [`mkdir -p`](https://github.com/postmodern/ruby-install/blob/61a627e436601c847cfa850c76cd3433d5e465b6/share/ruby-install/functions.sh#L11) which recursively creates the directories.

Mizou04: it won't create them if we don't have write permissions over `/this/part/exists/`

test/cli-tests/warn_if_no_write_permissions_test.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+. ./test/helper.sh
+
+test_install_dir="$test_fixtures_dir/no_write_permissions_test"
```
postmodern, line 5: Maybe to make this test more explicit, you could add another variable called `non_writable_parent_dir="$test_fixtures_dir/..."` and set `test_install_dir="$non_writable_parent_dir/ruby"`.

postmodern (changes requested):
Noticed a few issues with the code. I also still think the `while` loop can be replaced with `mkdir -p`.

bin/ruby-install
postmodern: If the `mkdir -p` fails because the last existing sub-directory isn't writable then the command will fail, which could be detected, an error message printed, and an error code returned.

PR #467, opened 2023-09-19
Updates for TruffleRuby and TruffleRuby GraalVM 23.1.0

eregon:
https://github.com/oracle/truffleruby/releases/tag/graal-23.1.0

For the Native Standalone the URL is slightly different, vm- => graal- for the tag name.

GraalVM no longer includes `gu` to install TruffleRuby in it, instead there is now a JVM Standalone, which is installable exactly like the Native Standalone.

### DIFF
--- a/share/ruby-install/truffleruby-graalvm/functions.sh
+++ b/share/ruby-install/truffleruby-graalvm/functions.sh
@@ -13,6 +13,8 @@ case "$os_arch" in
 	*)       fail "Unsupported platform $os_arch" ;;
 esac
 
+truffleruby_major="${ruby_version%%.*}"
+
 if [ "$ruby_version" = "23.0.0" ]; then
 	log "TruffleRuby-GraalVM 23.0 and later installed by ruby-install use the faster Oracle GraalVM distribution"
 	log "Oracle GraalVM uses the GFTC license, which is free for development and production use, see https://medium.com/graalvm/161527df3d76"
@@ -20,7 +22,7 @@ if [ "$ruby_version" = "23.0.0" ]; then
 	ruby_archive="${ruby_archive:-graalvm-jdk-17.0.7_${graalvm_platform/darwin/macos}-${graalvm_arch/amd64/x64}_bin.tar.gz}"
 	ruby_mirror="${ruby_mirror:-https://download.oracle.com/graalvm/17/archive}"
 	ruby_url="${ruby_url:-$ruby_mirror/$ruby_archive}"
-elif [ "${ruby_version%%.*}" -ge "23" ]; then # 23.1+
+elif (( truffleruby_major >= 23 )); then # 23.1+
 	ruby_dir_name="truffleruby-$ruby_version-${graalvm_platform/darwin/macos}-$graalvm_arch"
 	ruby_archive="${ruby_archive:-truffleruby-jvm-$ruby_version-${graalvm_platform/darwin/macos}-$graalvm_arch.tar.gz}"
 	ruby_mirror="${ruby_mirror:-https://github.com/oracle/truffleruby/releases/download}"
@@ -43,7 +45,7 @@ function install_ruby()
 	fi
 
 	log "Installing TruffleRuby GraalVM $ruby_version ..."
-	if [ "${ruby_version%%.*}" -ge "23" ]; then # 23.1+
+	if (( truffleruby_major >= 23 )); then # 23.1+
 		copy_into "$ruby_build_dir" "$install_dir" || return $?
 	else
 		copy_into "$ruby_build_dir" "$install_dir/graalvm" || return $?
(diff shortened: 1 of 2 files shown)


share/ruby-install/truffleruby-graalvm/functions.sh
```diff
@@ -20,6 +20,11 @@ if [ "$ruby_version" = "23.0.0" ]; then
 	ruby_archive="${ruby_archive:-graalvm-jdk-17.0.7_${graalvm_platform/darwin/macos}-${graalvm_arch/amd64/x64}_bin.tar.gz}"
 	ruby_mirror="${ruby_mirror:-https://download.oracle.com/graalvm/17/archive}"
 	ruby_url="${ruby_url:-$ruby_mirror/$ruby_archive}"
+elif [ "${ruby_version%%.*}" -ge "23" ]; then # 23.1+
```
postmodern, line 23: As a general rule-of-thumb, you should use `[[ ]]` for String comparisons and `(( ))` for arithmetic comparisons.
https://mywiki.wooledge.org/BashFAQ/031

Another benefit of `(( ))` over `[[ ]]`, is you can use regular `>` and `<` comparison operators instead of `-gt` or `-lt`.

share/ruby-install/truffleruby/functions.sh
```diff
@@ -29,6 +29,9 @@ if [ "$ruby_version" = "23.0.0" ]; then
 		*)              fail "Unsupported platform $truffleruby_platform-$truffleruby_arch" ;;
 	esac
 	ruby_url="${ruby_url:-$ruby_mirror/$truffleruby_artifact_id/content}"
+elif [ "${ruby_version%%.*}" -ge "23" ]; then # 23.1+
```
postmodern, line 32: Use `(( ))`. https://mywiki.wooledge.org/BashFAQ/031

share/ruby-install/truffleruby-graalvm/functions.sh
```diff
@@ -37,37 +42,46 @@ function install_ruby()
...
+	fi
 }
 
 #
 # Post-install tasks.
 #
 function post_install()
 {
-	cd "$install_dir/graalvm" || return $?
+	if [ "${ruby_version%%.*}" -ge "23" ]; then # 23.1+
```
postmodern, line 58: Use `(( ))`. https://mywiki.wooledge.org/BashFAQ/031

share/ruby-install/truffleruby-graalvm/functions.sh
```diff
@@ -37,37 +42,46 @@ function install_ruby()
 		return 1
 	fi
 
-	log "Installing GraalVM $ruby_version ..."
-	copy_into "$ruby_build_dir" "$install_dir/graalvm" || return $?
+	log "Installing TruffleRuby GraalVM $ruby_version ..."
+	if [ "${ruby_version%%.*}" -ge "23" ]; then # 23.1+
```
postmodern, line 46: Use `(( ))`. https://mywiki.wooledge.org/BashFAQ/031

postmodern (changes requested):
As a general rule-of-thumb, use `(( ))` instead of `[ ]` or `[[ ]]` for when you want to do integer comparisons, which allows you to use the `>` and `<` comparison operators instead of `-gt`, `-lt`.
https://mywiki.wooledge.org/BashFAQ/031

share/ruby-install/truffleruby-graalvm/functions.sh
eregon: I added a `truffleruby_major` variable, that feels cleaner and more reliable.

share/ruby-install/truffleruby-graalvm/functions.sh
```diff
@@ -37,37 +44,46 @@ function install_ruby()
 		return 1
 	fi
 
-	log "Installing GraalVM $ruby_version ..."
-	copy_into "$ruby_build_dir" "$install_dir/graalvm" || return $?
+	log "Installing TruffleRuby GraalVM $ruby_version ..."
+	if (( truffleruby_major >= 23 )); then # 23.1+
```
eregon, line 48: Actually this `if` will be taken for version `23.0.0` which is not wanted.
Any idea how I could get the minor component of the version in Bash?

eregon: I found a way

share/ruby-install/truffleruby-graalvm/functions.sh
postmodern: You could extract `${ruby_version%%.*}` into another variable to make it more explicit. `(( ))` is used when you want to "interpret" the variable's value as an integer.

postmodern (approved):

postmodern:
I will release 0.9.2 containing this, ahead of 0.10.0.

postmodern:
0.9.2 has been released, although it takes time for homebrew to update their formula.

Issue #458, opened 2023-07-05
`--with-opt-dir` doesn't prevent Ruby from seeing OpenSSL 3 headers/lib


dgholz:
## Description

This _does_ also happen when compiling Ruby manually, but I see in the [Ruby MacOS CI steps](https://github.com/ruby/ruby/blob/833b32f503abb9712dbb3a6825b1730c8ee1e912/.github/workflows/macos.yml#L67) that they use `--with-openssl-dir` instead.

Recently, [Homebrew updated loads of formulæ to use OpenSSL 3](https://github.com/Homebrew/homebrew-core/issues/134251), and I see the latest version of their OpenSSL package (`3.1.1_1`) get installed when I updated some frequently-used tools (like `curl`) . When I tried compiling Ruby afterwards, I was surprised to see some OpenSSL 3 errors appear in the output.

I traced it to how Ruby figures out where to find OpenSSL, and how the paths in the `--with-opt-dir` don't preclude it from making its own search. It then configures `~/src/ruby-3.1.4/ext/openssl/Makefile` like:
```
[...]
INCFLAGS = -I. -I$(arch_hdrdir) -I$(hdrdir) -I$(srcdir) -I/opt/homebrew/Cellar/openssl@3/3.1.1_1/include
DEFS     =
CPPFLAGS = -DRUBY_EXTCONF_H=\"$(RUBY_EXTCONF_H)\" -I/opt/homebrew/opt/openssl@1.1/include -I/opt/homebrew/opt/readline/include -I/opt/homebrew/opt/libyaml/include -I/opt/homebrew/opt/gdbm/include -D_XOPEN_SOURCE -D_DARWIN_C_SOURCE -D_DARWIN_UNLIMITED_SELECT -D_REENTRANT $(DEFS) $(cppflags)
CXXFLAGS = $(CCDLFLAGS) -fdeclspec  $(ARCH_FLAG)
ldflags  = -L. -fstack-protector-strong -L/opt/homebrew/opt/openssl@1.1/lib -L/opt/homebrew/opt/readline/lib -L/opt/homebrew/opt/libyaml/lib -L/opt/homebrew/opt/gdbm/lib -L/opt/homebrew/Cellar/openssl@3/3.1.1_1/lib
[...]
```

I noticed that specifying the path to OpenSSL explicitly let it work just fine:
```
$ ruby-install 3.1.4 -- --with-openssl-dir="$(brew --prefix openssl@1.1)"
[...]
[...]

Issue #461, opened 2023-08-28
Use OpenSSL 3 by default


semaperepelitsa:
With Ruby 3.1.4 I would get the following error when trying to install latest "puma" gem.

<details>

```
Installing puma 6.3.1 with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /Users/sema/.gem/ruby/3.1.0/gems/puma-6.3.1/ext/puma_http11
/Users/sema/.rubies/ruby-3.1.4/bin/ruby -I /Users/sema/.rubies/ruby-3.1.4/lib/ruby/3.1.0 extconf.rb
using OpenSSL pkgconfig (openssl.pc)
checking for openssl/bio.h... yes
checking for DTLS_method() in openssl/ssl.h... yes
checking for SSL_CTX_set_session_cache_mode(NULL, 0) in openssl/ssl.h... yes
checking for TLS_server_method() in openssl/ssl.h... yes
checking for SSL_CTX_set_min_proto_version(NULL, 0) in openssl/ssl.h... yes
checking for X509_STORE_up_ref()... yes
checking for SSL_CTX_set_ecdh_auto(NULL, 0) in openssl/ssl.h... yes
checking for SSL_CTX_set_dh_auto(NULL, 0) in openssl/ssl.h... yes
checking for SSL_get1_peer_certificate() in openssl/ssl.h... no
checking for Random.bytes... yes
creating Makefile

current directory: /Users/sema/.gem/ruby/3.1.0/gems/puma-6.3.1/ext/puma_http11
make DESTDIR\= sitearchdir\=./.gem.20230829-62642-3qoc84 sitelibdir\=./.gem.20230829-62642-3qoc84 clean

current directory: /Users/sema/.gem/ruby/3.1.0/gems/puma-6.3.1/ext/puma_http11
make DESTDIR\= sitearchdir\=./.gem.20230829-62642-3qoc84 sitelibdir\=./.gem.20230829-62642-3qoc84
compiling http11_parser.c
compiling mini_ssl.c
compiling puma_http11.c
linking shared-object puma/puma_http11.bundle
Undefined symbols for architecture arm64:
  "_SSL_get1_peer_certificate", referenced from:
      _engine_peercert in mini_ssl.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [puma_http11.bundle] Error 1

make failed, exit code 2
```

</details>

[...]

PR #468, opened 2023-09-27
Prefer --with-openssl-dir

k0kubun:
Fix https://github.com/postmodern/ruby-install/issues/458

When your homebrew has both `openssl@1.1` and `openssl@3` and you pass `--with-opt-dir=/opt/homebrew/opt/openssl@1.1` to `configure`, the build somehow mixes up both versions and fails.

It looks like `--with-openssl-dir` is the only correct way to specify an OpenSSL directory https://github.com/postmodern/ruby-install/issues/458#issuecomment-1667744262, so this PR uses it for Ruby 3.1.1 and newer (ref: https://github.com/postmodern/ruby-install/issues/458#issuecomment-1667746749). Older versions use the same option for backward compatibility, which would hopefully work if you don't have `openssl@3` locally.

### DIFF
--- a/share/ruby-install/ruby/functions.sh
+++ b/share/ruby-install/ruby/functions.sh
@@ -24,11 +24,13 @@ function configure_ruby()
 	fi
 
 	local opt_dir
+	local openssl_dir
 
 	log "Configuring ruby $ruby_version ..."
 	case "$package_manager" in
 		brew)
-			opt_dir="$(brew --prefix openssl@1.1):$(brew --prefix readline):$(brew --prefix libyaml):$(brew --prefix gdbm)"
+			opt_dir="$(brew --prefix readline):$(brew --prefix libyaml):$(brew --prefix gdbm)"
+			openssl_dir="$(brew --prefix openssl@1.1)"
 			;;
 		port)
 			opt_dir="/opt/local"
@@ -37,6 +39,7 @@ function configure_ruby()
 
 	run ./configure --prefix="$install_dir" \
 			"${opt_dir:+--with-opt-dir="$opt_dir"}" \
+			"${openssl_dir:+--with-openssl-dir="$openssl_dir"}" \
 			"${configure_opts[@]}" || return $?
 }
 


postmodern:
In issue #461, we're considering just switching to OpenSSL 3.x as the default. Most other Linux distributions have already switched to openssl 3.x and do not provide a legacy openssl 1.1 package. Seeing as how Ruby 3.0 will reach End-Of-Life March 31st, 2024, should we even bother supporting older Ruby versions?

k0kubun:
> should we even bother supporting older Ruby versions?

I changed the patch to always use `--with-openssl-dir`. I'll leave it to you to decide whether you merge this first or just do this in your OpenSSL 3 change together. As long as `--with-openssl-dir` will be used at some point, I'd be happy.

postmodern:
@k0kubun just to get clarification, can the openssl installation directory be specified using `--with-opt-dir`, or must it be specified using `--with-openssl-dir`? Should `ruby-install` replace usage of `--with-opt-dir` with individual `--with-*-dir` options for each Ruby dependency (ex: `--with-openssl-dir=... --with-yaml-dir=... ...`)?

k0kubun:
> can the openssl installation directory be specified using --with-opt-dir, or must it be specified using --with-openssl-dir?

I believe it must be `--with-openssl-dir`. On the openssl.gem side, it's the only documented way to build it https://github.com/ruby/openssl#installation. Specifying an OpenSSL directory with `--with-opt-dir` seems to partially work, but the rest of it relies on something not controlled by `--with-opt-dir`.

> Should ruby-install replace usage of --with-opt-dir with individual --with-*-dir options for each Ruby dependency (ex: --with-openssl-dir=... --with-yaml-dir=... ...)?

TBH I'm not sure what's the best practice for `libyaml` and `gdbm`. I've been able to build Ruby without specifying anything special for those things while I've often needed explicit `--with-openssl-dir` on macOS with homebrew for fixing build issues. I'd say it's fine to leave those two as is unless we know it's breaking something.

postmodern (approved):

Issue #429, opened 2022-08-03
ruby-install depends on xzcat, which is not available out of the box on all platforms


monfresh:
## Description

I maintain a script that installs Ruby on a Mac. If Homebrew is not yet present on the machine, the script installs Homebrew, then installs chruby and ruby-install, then installs the latest Ruby with ruby-install. Up until the latest `xzcat` changes in 0.8.4, it was working perfectly. But now, on a machine that didn't have Homebrew prior to running my script, the Ruby compilation fails with:

```
xzcat: command not found
```
Instead of assuming `xzcat` is in the PATH, would it be better to use the full path using the brew prefix, like `$(brew --prefix)/xzcat`?

cc @jmettraux

monfresh:
I originally thought that xzcat was installed by Homebrew because when I do `which xzcat`, it shows me `/usr/local/bin/xzcat` on Intel Macs and `/opt/homebrew/bin/xzcat` on Apple Silicon Macs. However, it seems like that might only be the case on certain macOS versions. I have an Intel Mac on Big Sur, and after installing Homebrew and restarting the terminal, xzcat is still not found. So, I don't think it's a good idea to assume it's available everywhere.

jmettraux:
> So, I don't think it's a good idea to assume it's available everywhere.

Since you install Homebrew, can you `brew install xzcat` right before installing ruby-install?

If it's too much of a hassle, I'm totally OK with my change being reverted, since OpenBSD usage compared to MacOSX usage is tiny.

monfresh:
OK, I found the issue. The reason why `xzcat` was available on my other machines, but not the Big Sur one is because I had somehow installed `xz` with Homebrew, probably as a dependency of another tool I installed. So, I can confirm that `xzcat` is NOT available by default on a Mac.

As for my script, I can fix this by adding `xz` (not `xzcat` because that formula doesn't exist on Homebrew) to my script's Brewfile.

jmettraux:
https://github.com/postmodern/ruby-install/blob/master/.github/workflows/shunit2.yml

postmodern:
We can add `xz` as a dependency to the homebrew formula.

monfresh:
@postmodern Yes, please. That seems like the a great solution.

postmodern:
[...]

PR #482, opened 2024-06-14
Install xz as regular dependency

kattouf:
I found a corner case when [previous](https://github.com/postmodern/ruby-install/issues/429) `xz` install on macOS implementation doesn't work and propose a solution.

The previous implementation covered cases when we installed `ruby-install` on macOS via `brew` and that worked well. 
But in our case, we install (implicitly) `ruby-install` using [mise](https://mise.jdx.dev/lang/ruby.html), that don't use brew to install it under the hood, and as a consequence we miss brew formula dependencies installation.

I propose to install `xz` as part of installing all other dependencies (via system package managers) that cover all cases.

### DIFF
--- a/homebrew/ruby-install.rb
+++ b/homebrew/ruby-install.rb
@@ -6,8 +6,6 @@ class RubyInstall < Formula
   license "MIT"
   head "https://github.com/postmodern/ruby-install.git", branch: "master"
 
-  depends_on 'xz'
-
   bottle do
     sha256 cellar: :any_skip_relocation, arm64_big_sur: "d19d4a89f08d447d522387e362b5e4627de5438e06396e2035fe735dc192d7de"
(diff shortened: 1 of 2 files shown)


postmodern (approved):

postmodern:
@wei18 could you open a separate issue for this. Please include additional information, such as whether you are using Homebrew or MacPorts, whether the package manager actually did install `xz`. Also, check if your package manager's installation `bin` directories are correctly added to your `$PATH`. If this is more of an issue about `mise install` not being able to run `brew install`/`ports install`, submit an issue to [Mise](https://github.com/jdx/mise) instead.

[ruby-install 0.10.1](https://github.com/postmodern/ruby-install/blob/v0.10.1/share/ruby-install/ruby/dependencies.sh#L61) should always be installing the `xz` package by default.

PR #492, opened 2025-01-10
Fix RPM build on COPR

gschlager:
It looks like all builds in existing Fedora COPR repositories are either failing or out-of-date. I created a new one at https://copr.fedorainfracloud.org/coprs/gschlager/ruby/ which I intend to keep up-to-date.

But I had to make changes to `rpm/ruby-install.spec` and delete `rpm/sources` to make the build work.

### DIFF
--- a/rpm/ruby-install.spec
+++ b/rpm/ruby-install.spec
@@ -14,6 +14,7 @@ License: MIT
 URL: https://github.com/postmodern/ruby-install#readme
 AutoReqProv: no
 BuildArch: noarch
+BuildRequires: make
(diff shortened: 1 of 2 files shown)


gschlager:
I'm not sure about the `sources` file. All I know is that the build failed because of it. It complained about the `c24c12b179ce197e38b72ec2c80ef93e` entry in the file.

postmodern:
Going to close this because we should not remove the `sources` checksum verification file; unless it's being replaced with something better. In the future, I'd want to setup GitHub Actions to automatically build an RPM for new releases and push it to an official RPM repository.

PR #505, opened 2025-02-18
Allow to install legacy Rubies with old OpenSSL versions on Mac/Homebrew

meineerde:
This fixes two issues which prevented the use of older Rubies:

* When requiring the use of OpenSSL < 3 with MacOS / Homebrew, we use the [rbenv/homebrew-tap](https://github.com/rbenv/homebrew-tap). This is required because Homebrew [has disabled](https://formulae.brew.sh/formula/openssl@1.1) the `openssl@1.1` formula, preventing its use by default.
* Ruby before 2.4 required OpenSSL 1.0. We confiogure the rbenv tap for openssl@1.0 for these versions.

With these patches, I was able to install Ruby at least down to 2.3.7 on an M4 Mac.

### DIFF
--- a/share/ruby-install/ruby/dependencies.sh
+++ b/share/ruby-install/ruby/dependencies.sh
@@ -98,6 +98,7 @@ fi
 case "$package_manager" in
 	brew|port)
 		case "$ruby_version" in
+			1.8|2.[0-3])	openssl_version="1.0" ;;
 			2.*|3.0.*)	openssl_version="1.1" ;;
 			*)		openssl_version="3" ;;
 		esac
@@ -109,7 +110,11 @@ esac
 # only for homebrew.
 #
 case "$package_manager" in
-	brew)	ruby_dependencies+=("openssl@${openssl_version}") ;;
+	brew)
+		openssl_package="openssl@${openssl_version}"
+		[[ "${openssl_version}" == 1.* ]] && openssl_package="rbenv/tap/${openssl_package}"
+		ruby_dependencies+=("${openssl_package}")
+		;;
 	port)	ruby_dependencies+=("openssl${openssl_version/./}") ;;
 esac
 


postmodern (commented):
I'm not sure how I feel about making `ruby-install` dependent on rbenv's custom homebrew packages. Also, I'm worried about the security implications of installing and building against an obviously old and unmaintained version of openssl.

If we are going to go ahead with this change, this needs tests! See the existing tests for how `ruby_dependencies` is populated in `test/ruby-tests/dependencies_tests.sh`.

postmodern:
After some discussion on [Mastodon](https://ruby.social), I've come to the conclusion we should not make it easy to install EoLed versions of Ruby. There should be an incentive to upgrade to a new supported version, and a disincentive to continue using an older EoLed version, which may also require specific versions of openssl that are also unsupported/EoLed.

meineerde:
I haven't seen [the discussion](https://ruby.social/@postmodern/114026968811273244) in time.

As [proposed there](https://toot.berlin/@holger/114036675975213720), we would add a warning when trying to install an old version in order to disincentive the using of older Rubies. If that is something you would consider, I'd try to take a stab at that.

In any case though, I still strongly believe that it should be possible to install older Rubies, even if they may be EOL'd upstream.

As for relying on rbenv's OpenSSL packages: if required, I'd be happy to try to address any issues this may cause. In general, these packages seem to be unencumbered (and unpatched) OpenSSL versions and apart from being maintained by rbenv, there is nothing Ruby- (or ruby-version-manager-)specific about them.

meineerde:
(Note that I have updated my branch in the meantime to address the points of your original review, including an added test. As the pull request is currently closed, these updates are not shown here. See https://github.com/postmodern/ruby-install/compare/master...meineerde:ruby-install:feature/legacy_rubies for the full changes.)

PR #524, opened 2026-03-26
Fix archive extraction to ensure it always extracts to the given path

### LINKED ISSUE #522: `ruby-install truffleruby-graalvm-33.0.0` is broken
## Steps To Reproduce

```
$ bin/ruby-install truffleruby-graalvm-33.0.0 --prefix /tmp/truffleruby
>>> Installing truffleruby-graalvm 33.0.0 into /tmp/truffleruby ...
>>> Installing dependencies for truffleruby-graalvm 33.0.0 ...
Updating and loading repositories:
Repositories loaded.
Package "make-1:4.4.1-11.fc43.x86_64" is already installed.
Package "gcc-15.2.1-7.fc43.x86_64" is already installed.
Package "zlib-ng-compat-devel-2.3.3-2.fc43.x86_64" is already installed.
Package "ca-certificates-2025.2.80_v9.0.304-1.2.fc43.noarch" is already installed.

Nothing to do.
>>> Downloading https://github.com/truffleruby/truffleruby/releases/download/graal-33.0.0/truffleruby-jvm-33.0.0-linux-amd64.tar.gz into /home/eregon/src ...
HTTP response 302 ...[===========================================================================>]  152.96M   21.96MB/silename%3Dtruffleruby-jvm-3[Files: 1  Bytes: 152.96M [19.22MB/s] Redirects: 1  Todo: 0  Errors: 0       ]
>>> Verifying truffleruby-jvm-33.0.0-linux-amd64.tar.gz ...
>>> Extracting truffleruby-jvm-33.0.0-linux-amd64.tar.gz to /home/eregon/src/truffleruby-33.0.0-linux-amd64 ...
bin/ruby-install: line 53: cd: /home/eregon/src/truffleruby-33.0.0-linux-amd64: No such file or directory
!!! ruby build directory is missing: /home/eregon/src/truffleruby-33.0.0-linux-amd64
zsh: exit 255
```

Indeed there is no `src/truffleruby-33.0.0-linux-amd64`, instead there is:
```
$ ls -l ~/src
...
drwxr-xr-x.  9 eregon eregon      4096 Mar 26 14:20 truffleruby-jvm-33.0.0-linux-amd64
-rw-r--r--.  1 eregon eregon 160400569 Jan  9 12:46 truffleruby-jvm-33.0.0-linux-amd64.tar.gz
```

The issue here is the name of the folder inside the archive changed, from truffleruby-graalvm 25.0.0+ to include `jvm` and before it did not:
```
for f in truffleruby-jvm-*-macos-aarch64.tar.gz; do echo $f; tar tf $f | head -1; echo; done                                        
truffleruby-jvm-23.1.0-macos-aarch64.tar.gz
truffleruby-23.1.0-macos-aarch64/native-image.properties
zsh: broken pipe  tar tf $f | 
zsh: done         head -1

truffleruby-jvm-24.2.2-macos-aarch64.tar.gz
truffleruby-24.2.2-macos-aarch64/LICENSE_TRUFFLERUBY.txt
zsh: broken pipe  tar tf $f | 
zsh: done         head -1

truffleruby-jvm-25.0.0-macos-aarch64.tar.gz
truffleruby-jvm-25.0.0-macos-aarch64/native-image.properties
zsh: broken pipe  tar tf $f | 
zsh: done         head -1

truffleruby-jvm-33.0.1-macos-aarch64.tar.gz
truffleruby-jvm-33.0.1-macos-aarch64/native-image.properties
zsh: broken pipe  tar tf $f | 
zsh: done         head -1
```

This gets worse because it would accidentally "work wrongly/not fail" if one does:
```
$ bin/ruby-install truffleruby-33.0.0 --prefix /tmp/truffleruby-native
$ bin/ruby-install truffleruby-graalvm-33.0.0 --prefix /tmp/truffleruby
SUCCESS! BUT:
$ /tmp/truffleruby/bin/ruby -v
truffleruby 33.0.0 (2026-01-09), like ruby 3.3.7, Oracle GraalVM Native [x86_64-linux]
UH-OH that's wrong, should be Oracle GraalVM JVM
```

because the same folder `truffleruby-33.0.0-linux-amd64` gets reused.

Other Ruby installers do not have this problem.
The fundamental issue here is `ruby-install` does not take much precautions that the archive extracts as the expected name, and requires hardcoding the name of the folder inside the archive which is (as proven by this issue) brittle.

I will make a PR to ensure ruby-install always extract to the proper place without conflicts and no longer needs hardcoded folder names inside the archive.

Found by @nirvdrum

eregon:
* Fixes https://github.com/postmodern/ruby-install/issues/522
* The various functions.sh no longer need to hardcode $ruby_dir_name.
* Remove $ruby_dir_name entirely as it is now unused.

### DIFF
--- a/share/ruby-install/util.sh
+++ b/share/ruby-install/util.sh
@@ -74,29 +74,39 @@ function download()
 function extract()
 {
 	local archive="$1"
-	local dest="${2:-${archive%/*}}"
+	local dest="$2"
+	local tmp="$dest.tmp"
 
-	mkdir -p "$dest" || return $?
+	rm -rf "$dest" "$tmp" || return $?
+	mkdir -p "$tmp" || return $?
 
 	case "$archive" in
 		*.tgz|*.tar.gz)
-			run tar -xzf "$archive" -C "$dest" || return $?
+			run tar -xzf "$archive" -C "$tmp" || return $?
 			;;
 		*.tbz|*.tbz2|*.tar.bz2)
-			run tar -xjf "$archive" -C "$dest" || return $?
+			run tar -xjf "$archive" -C "$tmp" || return $?
 			;;
 		*.txz|*.tar.xz)
-			debug "xzcat $archive | tar -xf - -C $dest"
-			xzcat "$archive" | tar -xf - -C "$dest" || return $?
+			debug "xzcat $archive | tar -xf - -C $tmp"
+			xzcat "$archive" | tar -xf - -C "$tmp" || return $?
 			;;
 		*.zip)
-			run unzip "$archive" -d "$dest" || return $?
+			run unzip "$archive" -d "$tmp" || return $?
 			;;
 		*)
 			error "Unknown archive format: $archive"
 			return 1
 			;;
 	esac
+
+	local extracted=("$tmp"/*)
+	if (( ${#extracted[@]} != 1 )); then
+		error "Multiple extracted directories under $tmp"
+		return 1
+	fi
+	mv "${extracted[0]}" "$dest" || return $?
+	rm -rf "$tmp"
 }
 
 #
--- a/share/ruby-install/functions.sh
+++ b/share/ruby-install/functions.sh
@@ -48,7 +48,7 @@ function verify_ruby()
 function extract_ruby()
 {
 	log "Extracting $ruby_archive to $ruby_build_dir ..."
-	extract "$src_dir/$ruby_archive" "$src_dir" || return $?
+	extract "$src_dir/$ruby_archive" "$ruby_build_dir" || return $?
 }
 
 #
--- a/share/ruby-install/jruby/functions.sh
+++ b/share/ruby-install/jruby/functions.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
 ruby_archive="${ruby_archive:-jruby-dist-$ruby_version-bin.tar.gz}"
-ruby_dir_name="jruby-$ruby_version"
 ruby_mirror="${ruby_mirror:-https://repo1.maven.org/maven2/org/jruby/jruby-dist}"
 ruby_url="${ruby_url:-$ruby_mirror/$ruby_version/$ruby_archive}"
 
--- a/share/ruby-install/mruby/functions.sh
+++ b/share/ruby-install/mruby/functions.sh
@@ -1,7 +1,6 @@
 #!/usr/bin/env bash
 
 ruby_archive="${ruby_archive:-mruby-$ruby_version.tar.gz}"
-ruby_dir_name="mruby-$ruby_version"
 ruby_mirror="${ruby_mirror:-https://github.com/mruby/mruby/archive}"
 ruby_url="${ruby_url:-$ruby_mirror/$ruby_version/$ruby_archive}"
 
--- a/share/ruby-install/ruby-install.sh
+++ b/share/ruby-install/ruby-install.sh
@@ -295,7 +295,7 @@ function init()
 
 	ruby_cache_dir="$ruby_install_cache_dir/$ruby"
 	install_dir="${install_dir:-$rubies_dir/$ruby-$ruby_version}"
-	ruby_build_dir="$src_dir/$ruby_dir_name"
+	ruby_build_dir="$src_dir/$ruby-$ruby_version"
 
 	ruby_md5="${ruby_md5:-$(ruby_checksum_for "$ruby" md5 "$ruby_archive")}"
 	ruby_sha1="${ruby_sha1:-$(ruby_checksum_for "$ruby" sha1 "$ruby_archive")}"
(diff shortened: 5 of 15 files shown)


eregon:
@postmodern Could you review this PR? It fixes a confusing bug: https://github.com/postmodern/ruby-install/issues/522

omarharby805-ship-it (approved):

share/ruby-install/util.sh
```diff
@@ -74,29 +74,39 @@ function download()
 function extract()
 {
 	local archive="$1"
-	local dest="${2:-${archive%/*}}"
+	local dest="$2"
+	local dest_dir="$(dirname "$2")"
+	local tmp="$dest_dir/tmp"
 
-	mkdir -p "$dest" || return $?
+	rm -rf "$dest"
```
havenwood: This should probably have a trailing `|| return $?` in case the `rm` fails. (Just to match project convention, and if it carries on after this `rm` fails the `mv` below might confusingly move into `$dest` instead of renaming.)

share/ruby-install/util.sh
```diff
@@ -74,29 +74,39 @@ function download()
...
 			return 1
 			;;
 	esac
+
+	local extracted=("$tmp"/*)
+	if (( ${#extracted[@]} != 1 )); then
+		error "Multiple extracted directories under $tmp"
+		return 1
+	fi
+	mv "${extracted[0]}" "$dest"
```
havenwood: It might be a good idea to add an `rm -rf "$tmp"` here to clean up the now-empty `$tmp` dir. (If an `rm` is added, the `mv` would need a ` || return $?`.)

share/ruby-install/util.sh
```diff
@@ -74,29 +74,39 @@ function download()
 function extract()
 {
 	local archive="$1"
-	local dest="${2:-${archive%/*}}"
+	local dest="$2"
+	local dest_dir="$(dirname "$2")"
```
havenwood: `dirname "$dest"` might be nice? `"$2"` and `"$dest"` are same here, so just an optional nit.

havenwood (changes requested):
This seems like a nice fix for #522. I left a couple suggestions around propagating errors up and clearing the temp dir, but LGTM!

share/ruby-install/util.sh
eregon: Good catch, `~/src/tmp` was leftover as an empty dir before this, done.

share/ruby-install/util.sh
eregon: Added

share/ruby-install/util.sh
eregon: Done

eregon:
Would love if this could get merged soon :)

havenwood (approved):
Thank you!

share/ruby-install/util.sh
```diff
@@ -74,29 +74,40 @@ function download()
 function extract()
 {
 	local archive="$1"
-	local dest="${2:-${archive%/*}}"
+	local dest="$2"
+	local dest_dir="$(dirname "$dest")"
+	local tmp="$dest_dir/tmp"
 
-	mkdir -p "$dest" || return $?
+	rm -rf "$dest" || return $?
```
havenwood:
```suggestion
	rm -rf "$dest" "$tmp" || return $?
```

havenwood: I think we probably should clear `$tmp` up front here too in case of an interrupted prior run.

share/ruby-install/util.sh
```diff
@@ -74,29 +74,40 @@ function download()
 function extract()
 {
 	local archive="$1"
-	local dest="${2:-${archive%/*}}"
+	local dest="$2"
+	local dest_dir="$(dirname "$dest")"
```
havenwood:
```suggestion
	local tmp="$dest.tmp"
```

havenwood: Thinking about this more, I'm worried `$dest_dir` basically being `$src_dir` will bork parallel installs of different Rubies trying to share a tmp dir.

`$dest` is unique so I wonder about following `download()`'s `"$dest.part"` pattern with `"$dest.tmp"` here.

havenwood (changes requested):

havenwood:
> Would love if this could get merged soon :)

After staring at the diff for a while I got worried about parallel installs and left a few more comments, but everything else looks good to me.

Sorry for not catching that on first review. Getting the tmp dir part right is just the part that gives me pause to merge while postmodern is busy.

share/ruby-install/util.sh
eregon: Done

share/ruby-install/util.sh
eregon: Good point, I should have thought about that.
That's an easy fix, done.
`mktemp` could also be used but might not always be available and feels more complicated.

eregon:
Thank you for the thorough review, all fixed now and I also re-tested extensively locally, all good.

eregon:
Could this important fix be merged?

havenwood:
> Could this important fix be merged?

Yes, I think we should merge. Postmodern is busy and I'd really like to test it thoroughly before release and haven't had a chance. Will do!

